PR #2, opened 2019-03-05
Update reader

FrancaCassol:
Update of the LST reader with the reader version and the containers present in lstchain at the moment.
When this PR will be accepted  the lstchain version can be removed

### DIFF
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -206,12 +206,11 @@ class LSTEventSource(EventSource):
         event_container.tib_tenMHz_counter = unpacked_tib[2]
         event_container.tib_stereo_pattern = unpacked_tib[3]
         event_container.tib_masked_trigger = unpacked_tib[4]
-        #event_container.cdts_data = event.lstcam.cdts_data
         event_container.swat_data = event.lstcam.swat_data
 
         # unpack CDTS data
         rec_fmt = '=IIIQQBBB'
-        unpacked_cdts =  struct.unpack(rec_fmt, event.lstcam.cdst_data)
+        unpacked_cdts =  struct.unpack(rec_fmt, event.lstcam.cdts_data)
         event_container.ucts_event_counter = unpacked_cdts[0]
         event_container.ucts_pps_counter = unpacked_cdts[1]
         event_container.ucts_clock_counter = unpacked_cdts[2]
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -5,9 +5,10 @@ EventSource for LSTCam protobuf-fits.fz-files.
 Needs protozfits v1.4.2 from github.com/cta-sst-1m/protozfitsreader
 """
 import numpy as np
-
+import struct
 from astropy import units as u
 import glob
+
 from ctapipe.core import Provenance
 from ctapipe.instrument import (
     TelescopeDescription,
@@ -17,22 +18,17 @@ from ctapipe.instrument import (
 )
 
 from ctapipe.io import EventSource
-from astropy.io import fits
 
-from .containers import LSTDataContainer
+from .containers import LSTDataContainer, PixelStatusContainer
 
 
 __all__ = ['LSTEventSource']
 
 
 class LSTEventSource(EventSource):
-
-    """
-    EventSource for LST r0 data.
-    """
+    """EventSource for LST r0 data."""
 
     def __init__(self, config=None, tool=None, **kwargs):
-
         """
         Constructor
         Parameters
@@ -54,6 +50,7 @@ class LSTEventSource(EventSource):
         # EventSource can not handle file wild cards as input_url
         # To overcome this we substitute the input_url with first file matching
         # the specified file mask (copied from  MAGICEventSourceROOT).
+
         if 'input_url' in kwargs.keys():
             self.file_list = glob.glob(kwargs['input_url'])
             self.file_list.sort()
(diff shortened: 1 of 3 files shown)


ctapipe_io_lst/__init__.py
```diff
@@ -184,51 +197,106 @@ def fill_lst_event_container_from_zfile(self, event):
...
+        event_container.tib_tenMHz_counter = unpacked_tib[2]
+        event_container.tib_stereo_pattern = unpacked_tib[3]
+        event_container.tib_masked_trigger = unpacked_tib[4]
+        #event_container.cdts_data = event.lstcam.cdts_data
         event_container.swat_data = event.lstcam.swat_data
-        event_container.counters = event.lstcam.counters
+
+        # unpack CDTS data
+        rec_fmt = '=IIIQQBBB'
+        rec_unpack = struct.Struct(rec_fmt).unpack_from
```
MaxNoe: why not simpy `struct.unpack(rec_fmt, event.lstcam.cdts_data)`

ctapipe_io_lst/__init__.py
```diff
@@ -184,51 +197,106 @@ def fill_lst_event_container_from_zfile(self, event):
...
+
+        event_container.pps_counter = np.zeros(self.camera_config.lstcam.num_modules)
+        event_container.tenMHz_counter = np.zeros(self.camera_config.lstcam.num_modules)
+        event_container.event_counter = np.zeros(self.camera_config.lstcam.num_modules)
+        event_container.trigger_counter = np.zeros(self.camera_config.lstcam.num_modules)
+        event_container.local_clock_counter = np.zeros(self.camera_config.lstcam.num_modules)
+        for mod in range(self.camera_config.lstcam.num_modules):
+
+            words=event.lstcam.counters[mod*rec_len:(mod+1)*rec_len]
```
MaxNoe: This looks like you could use `np.frombuffer` here to avoid this loop and fill the arrays directly

ctapipe_io_lst/__init__.py
FrancaCassol: No specific reason, if it makes the same I can change it

ctapipe_io_lst/__init__.py
FrancaCassol: Thank for the suggestion Max, unfortunately it is not immediate to me how to do, could you please show it to me?

MaxNoe: `counters` is just bytes, right? Bytes with a specific meaning. If i read the code correct, you have for each module, 

pps_counter, uint2
10 MHz couner,  uint4
event_counter uint4
trigger_counter uint4
local_clock_counter uint8

So yo load this immediatly into one array, you can do:
```
module_data = np.frombuffer(lstcam.counters, dtype=[('pps_counter', 'uint2'), ('counter_10Mhz', 'uint4'), ...])
```

MaxNoe: See https://docs.scipy.org/doc/numpy/reference/generated/numpy.frombuffer.html

MaxNoe:
Also, you should not specify native byteorder when reading from a file. Files are produced with a specific byte order, which you need to know to read them. FITS standard requires big-endian, but the custom binary blob stored in zfits files might store little endian, i don't know.

Why do you have to deal with binary data anyway? Should the zfits reader not give you arrays for each column directly?

FrancaCassol:
> 
> Why do you have to deal with binary data anyway? Should the zfits reader not give you arrays for each column directly?

Unfortunately not. But you are right I can try to ask this improvement to the Event Builder people ;-)

ctapipe_io_lst/__init__.py
FrancaCassol: Thanks Max, I will take into account the possibility to change as you suggest. But is it mandatory now? We are trying to put things all together and I would not change the code related to these counters so in hurry. If the code as it is now is acceptable, please accept the PR, I promise I will make the change soon

MaxNoe:
I do not need to accept anything here, this is basically your repository. It is not ctapipe code anymore. You do not have to wait for approved PRs. I was pnly making suggestions.

ctapipe_io_lst/__init__.py
MaxNoe: I  do not need to accept anything here, this is basically your repository. And right now, no
approved PR reviews are necessary to merge. 

I don't have the time and the in-depth knowledge of LST to properly review this code. 
There are to much LST internals involved.

 It is not ctapipe code anymore. You do not have to wait for approved  PRs. I was only making suggestions. I would however recommend, that those interested in LST reading, start reviewing here. So if there is anyone with enough knowledge of the LST internals, you should ask them if they could review.

FrancaCassol: I understand, I thank you for your help, it is very welcome (only I am a bit in hurry with this code right now and I know that people are using it and it works even if it can be surely be improved)
Still two questions:
- Who has the right to accept the merge? 
- Do you understand why one of the travis check does not work?

MaxNoe: * You can just merge it yourself
* Yes, it is because the tagged version of ctapipe I tested against, requires pyhessio which could not be installed. I will add it on the current master.

MaxNoe: As long as there is no group of people reviewing here, you could also directly push to the master branchf for smaller, simple changes.

MaxNoe:
So for now, we only test against the current master of ctapipe.

As soon as the next release of ctapipe is made, I would suggest to test both against the master and the latest release again.

FrancaCassol:
Ok, I will do it, thanks!

How can I finalize this PR now? I don't see where to do it ...

MaxNoe:
Ah sorry, you did not have write access. @mackaiver  added you to the @cta-observatory/cta-devs group, and I gave this group write access here.

Is this fine with you? Or should we only create an lst group with a few people that have write access?

FrancaCassol:
> Is this fine with you? Or should we only create an lst group with a few people that have write access?

Thanks Max, for the moment it is ok like this, I will discuss the question with the other LST developers  and let you know

PR #3, opened 2019-03-12
Improved nomenclature

FrancaCassol:
Improved PedestalContainer data naming

### DIFF
--- a/ctapipe_io_lst/containers.py
+++ b/ctapipe_io_lst/containers.py
@@ -31,53 +31,55 @@ class FlatFieldContainer(Container):
         unit=u.s
     )
     n_events = Field(0, 'Number of events used for statistics')
-    relative_gain_mean = Field(
-        None,
-        "np array of the relative flat-field coefficient mean (n_chan X N_pix)"
-    )
+
     charge_mean = Field(
         None,
-        "np array of signal charge mean (n_chan X N_pix)"
+        "np array of signal charge mean (n_chan X n_pix)"
     )
     charge_median = Field(
         None,
-        "np array of signal charge median (n_chan X N_pix)"
+        "np array of signal charge median (n_chan X n_pix)"
     )
-    charge_rms = Field(
+    charge_std = Field(
         None,
-        "np array of signal charge rms (n_chan X N_pix)"
+        "np array of signal charge standard deviation (n_chan X n_pix)"
     )
     time_mean = Field(
         None,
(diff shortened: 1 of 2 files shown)


ctapipe_io_lst/containers.py
```diff
@@ -93,16 +93,15 @@ class PedestalContainer(Container):
...
         None,
         "np array of pedestal average (n_chan X N_pix)"
     )
-    pedestal_median = Field(
+    charge_median = Field(
         None,
         "np array of the pedestal  median (n_chan X N_pix)"
     )
-    pedestal_rms = Field(
+    charge_rms = Field(
```
MaxNoe: Is this really the RMS? 

Or is it what physicists regularly call the RMS, but actually is the standard deviation?

So is it calculated as 

sqrt(<x²> - <x>²) or sqrt(<x²>) ?

FrancaCassol: Hi Max,

RMS  (as in high energy physics ;-) ) is intended to be the "root mean square deviation": https://en.wikipedia.org/wiki/Root-mean-square_deviation

for calculating it I am using at present the standard deviation from  numpy

std = sqrt(mean(abs(x - x.mean())**2)

Do you see a problem here?

MaxNoe: Yes. 

We should not  carry over errors made in PAW in the eighties, that were taken over into ROOT because of "backwards compatibility" into the next generation of frameworks.

RMS = Root Mean Square = sqrt(E(x²)) (https://en.wikipedia.org/wiki/Root_mean_square).

If you calculate the standard devition, please call it standard deviation.

MaxNoe: See: https://root.cern/root/html534/TH1.html#TH1:GetRMS

>  Note that this function returns the Standard Deviation (Sigma)
  of the distribution (not RMS).
  The name "RMS" was introduced many years ago (Hbook/PAW times).
  We kept the name for continuity.

FrancaCassol: Ok, I can proceed with this change, this will ask also a new PR in
https://github.com/cta-observatory/cta-lstchain/pull/92

FrancaCassol:
Sorry I just see an inconsistency in the initialisation of the PixelStatusContainer, I must correct it and do a new PR before merging

FrancaCassol:
If there are not objections I am going to merge this PR

PR #8, opened 2019-05-29
Make reader compatible with crappie

FrancaCassol:
This PR update the reader to the present version of ctapipe which now contains the calibration containers.

### DIFF
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -78,6 +78,7 @@ class LSTEventSource(EventSource):
         self.data = LSTDataContainer()
         self.data.meta['input_url'] = self.input_url
         self.data.meta['max_events'] = self.max_events
+        self.data.meta['origin'] = 'LSTCAM'
 
         # fill LST data from the CameraConfig table
         self.fill_lst_service_container_from_zfile()
@@ -321,13 +322,12 @@ class LSTEventSource(EventSource):
 
         # initialize the container
         status_container = PixelStatusContainer()
-        status_container.hardware_mask = np.zeros([self.n_camera_pixels], dtype=bool)
-        status_container.pedestal_mask = np.zeros([self.n_camera_pixels], dtype=bool)
-        status_container.flatfield_mask = np.zeros([self.n_camera_pixels], dtype=bool)
+        status_container.hardware_failing_pixels = np.zeros([self.n_camera_pixels], dtype=bool)
+        status_container.pedestal_failing_pixels = np.zeros([self.n_camera_pixels], dtype=bool)
+        status_container.flatfield_failing_pixels = np.zeros([self.n_camera_pixels], dtype=bool)
 
         mon_camera_container.pixel_status = status_container
         
-
     def fill_mon_container_from_zfile(self, event):
         """
         Fill with MonitoringContainer.
@@ -343,9 +343,7 @@ class LSTEventSource(EventSource):
             event.pixel_status
 
         # initialize the hardware mask
-        status_container.hardware_mask = pixel_status == 0
-
-
+        status_container.hardware_failing_pixels = pixel_status == 0
 
 
 class MultiFiles:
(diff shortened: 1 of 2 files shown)


ctapipe_io_lst/__init__.py
```diff
@@ -18,16 +18,26 @@
...
 class LSTEventSource(EventSource):
     """EventSource for LST r0 data."""
 
+    n_gains = Int(
+        2,
+        help='Number of gains at r0/r1 level'
+    ).tag(config=True)
+
+    baseline = Int(
+        250,
```
vuillaut: Just by curiosity, why 250?

FrancaCassol: This is actually the value I used for NectarCAM, in the case of the LST EvB3 should be 400 (I just changed it), but actually now (with EvB2) we are not using this value because the baseline subtraction is done at the r0 calibration level

ctapipe_io_lst/__init__.py
```diff
@@ -18,16 +18,26 @@
...
     """EventSource for LST r0 data."""
 
+    n_gains = Int(
+        2,
+        help='Number of gains at r0/r1 level'
+    ).tag(config=True)
+
+    baseline = Int(
+        400,
+        help='r0 waveform baseline '
```
rlopezcoto: Hi Franca, the LST EvB2 data, to which no correction has been applied has a baseline of 300, right?

ctapipe_io_lst/containers.py
```diff
@@ -22,185 +16,6 @@
...
-    Root container for monitoring data (MON)
-    """
-
-    tels_with_data = Field([], "list of telescopes with data")
-
-    # create the camera container
-    tel = Field(
-        Map(MonitoringCameraContainer),
-        "map of tel_id to MonitoringCameraContainer")
-
```
rlopezcoto: Just to keep track of things, all these containers are the ones that have been moved to `ctapipe/calib/camera`

ctapipe_io_lst/__init__.py
FrancaCassol: Now the baseline is defined by Pawel's code and it is 300, but in the future this will be set by the EvB3 and Julien Houle said to me that he will use a baseline at 400 (hardcoded in the EvB), 
if he will change his mind we can always change this value (I said it is not used for the moment)

ctapipe_io_lst/containers.py
FrancaCassol: yes, they are all in ctapipe now

FrancaCassol:
I think I solved the problem of the multiple file reading, but I want to test it a bit more (and now I have to leave). I will commit the code to this PR as soon as possible

ctapipe_io_lst/__init__.py
```diff
@@ -51,15 +69,36 @@ def __init__(self, **kwargs):
...
+
+        super().__init__(**kwargs)
+
+        if self.multi_streams:
+            # test how many streams are there:
+            # file name must be [stream name]Run[all the rest]
+            # All the files with the same [all the rest] are opened
+
+            if '/' in self.input_url:
+                dir, name = self.input_url.rsplit('/', 1)
```
MaxNoe: Please us `os.path.split`  here

MaxNoe: It will also work  if there is no `/` in the path. Also you could combine it with `os.path.abspath`, so you don't need any case differentiation here.

```
directory, filename = os.path.split(os.path.abspath(self.input_url))
```

ctapipe_io_lst/__init__.py
```diff
@@ -51,15 +69,36 @@ def __init__(self, **kwargs):
...
+                stream, run = name.split('Run', 1)
+            else:
+                run = name
+
+            ls = listdir(dir)
+            self.file_list = []
+
+            for file_name in ls:
+                if run in file_name:
+                    full_name = dir + '/' + file_name
```
MaxNoe: `os.path.join`

PR #10, opened 2019-08-01
Adding LSTDriveContainer

labsaha:
This PR is following a discussion on Drive reports and pointing positions at `cta-lstchain` https://github.com/cta-observatory/cta-lstchain/pull/153 with @FrancaCassol, @vuillaut and others.
@FrancaCassol Please have a look here if this is the way you want this in `LSTMonitoringContainer`. If not, suggest possible changes.


labsaha:
@FrancaCassol Thanks. I'll update this as you suggested.

PR #12, opened 2019-10-29
fix ctapipe version to 0.7 and update to use last ctapipe-extra

vuillaut:


MaxNoe:
Conda **env** create

vuillaut:
> Conda **env** create

*#!@#$%@!!
 thanks!
This build must actually never have worked...

MaxNoe:
No, `conda create` creates a new env, but does not support creating it from file. That's only supported by `conda env create`.

:man_shrugging: :monkey:

PR #15, opened 2019-11-23
Add criteria to decide from where to take timestamp and trigger valeus

FrancaCassol:
In this PR I add the check of the flag that extdevices_presence flag (discovered only recently) that tells which devices are seen by the EvB. The R0  time and trigger are set with the following criteria: if UTCS is available its values are taken, otherwise the TIB values.

The timestamp is for the moment in ns (is it ok?)

I add also the unpacking go the SWAT data, even if they are empty of the moment.

### DIFF
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -8,8 +8,9 @@ import numpy as np
 import struct
 from astropy import units as u
 import os
-from os import listdir, getcwd
-
+from os import listdir
+from astropy.time import Time
+from datetime import datetime
 from ctapipe.core import Provenance
 from ctapipe.instrument import (
     TelescopeDescription,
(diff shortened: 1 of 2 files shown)


morcuended:
Thanks @FrancaCassol! This information was missing in lstchain (see https://github.com/cta-observatory/cta-lstchain/issues/227)

> The timestamp is for the moment in ns (is it ok?) 

Yes, UCTS timestamp is in ns (UNIX format)

FrancaCassol:
Hi @morcuended, 
yes, this is precious information. You owe me a beer  ;-)

ctapipe_io_lst/__init__.py
```diff
@@ -281,19 +327,27 @@ def fill_lst_event_container_from_zfile(self, event):
...
-        # temporary patch to have an event time set
-        r0_container.trigger_time = (
-            self.data.lst.tel[self.camera_config.telescope_id].evt.tib_pps_counter +
-            self.data.lst.tel[self.camera_config.telescope_id].evt.tib_tenMHz_counter * 10**(-7))
+        if self.data.lst.tel[self.camera_config.telescope_id].evt.ucts_timestamp > 0:
+            r0_container.trigger_time = self.data.lst.tel[self.camera_config.telescope_id].evt.ucts_timestamp
 
-        if r0_container.trigger_time is None:
+        elif self.data.lst.tel[self.camera_config.telescope_id].evt.tib_pps_counter >0:
+            r0_container.trigger_time = (
```
morcuended: ...evt.ucts_timestamp is an absolute timestamp, on the contrary, if UCTS fails you define here trigger_time as the time that has elapsed since a starting ntp. Is that right? Either you define it as an absolute value or a value relative to a starting point.

morcuended: This is how I was taking the timing info in lstchain:

```
ucts_timestamp = event.lst.tel[telescope_id].evt.ucts_timestamp
 
if ucts_timestamp != 0:
    # UCTS timestamps are nanoseconds in UNIX timestamp format and must be given in seconds
    utc_time = Time(datetime.utcfromtimestamp(ucts_timestamp / 1e9))
else:
     start_ntp = event.lst.tel[telescope_id].svc.date
     ntp_time = start_ntp + event.r0.tel[telescope_id].trigger_time
     utc_time = Time(datetime.utcfromtimestamp(ntp_time))
```

morcuended:
> if UTCS is available its values are taken, otherwise the TIB values.

Does 'If UCTS timestamps are available' means that you can fully trust that their values are correct or just that the timestamps values are not null? Why trust more in UCTS than in TIB?

FrancaCassol:
> > if UTCS is available its values are taken, otherwise the TIB values.
> 
> Does 'If UCTS timestamps are available' means that you can fully trust that their values are correct or just that the timestamps values are not null? Why trust more in UCTS than in TIB?

Hi @morcuended, this is what several experts in a mail exchange suggested to do. But, obviously they suppose that the UCTS values are corrects, which seems not the case at the moment (probably  we should better use our time to make all pieces working correctly, one after the other and in the correct order, before trying to put all together without robust tests)

rlopezcoto:
Hi @FrancaCassol, @morcuended. If the UCTS is not meaningful, we should stick in this PR to using TIB times, even if we have to change it afterwards when it is fixed.

ctapipe_io_lst/__init__.py
labsaha: So it appears to me that it first checks if TIB info is present;  if so, then it will take TIB info and ignore other information. Therefore, it means TIB info will be given priority. If this is the case, then it is fine.

FrancaCassol:
> Hi @FrancaCassol, @morcuended. If the UCTS is not meaningful, we should stick in this PR to using TIB times, even if we have to change it afterwards when it is fixed.

Hi @ruben, 
yes, definitely this PR now is "outdated" ;-) (I did it thinking that UCTS is working)
I am going to change it giving priority to the TIB time but keeping  the rest of the code, in particular the check on the extdevice_presence

morcuended:
Even TIB times go crazy at some point after some subruns:
run1615_evid_time

But, better this than having those steps created by UCTS

FrancaCassol:
Hi, 
as it is clear now, things are generally not working properly. So, we should be simply careful in looking at the data and don't trust them blindly. I think we can in any case merge this and hope to have a better situation with all working properly soon

morcuended:
Yes, let's go ahead!

FrancaCassol:
Hi @morcuended, 
I think you should post your plot in the reco list, it is important to notify it

rlopezcoto:
In slack as well, thanks!

Issue #18, opened 2019-12-10
Change of formats to new DAQ Model version


morcuended:
While trying to process run1720, I got these errors:

```
Traceback (most recent call last):
...
  File "/fefs/home/daniel.morcuende/software/lstchain_v2_dev/cta-lstchain/lstchain/reco/dl0_to_dl1.py", line 247, in r0_to_dl1
    for i, event in enumerate(source):
  File "/home/daniel.morcuende/.conda/envs/lstchain_v2_dev/lib/python3.7/site-packages/ctapipe/io/eventsource.py", line 217, in __iter__
    for event in self._generator():
  File "/home/daniel.morcuende/.conda/envs/lstchain_v2_dev/lib/python3.7/site-packages/ctapipe_io_lst/__init__.py", line 153, in _generator
    self.fill_lst_event_container_from_zfile(event)
  File "/home/daniel.morcuende/.conda/envs/lstchain_v2_dev/lib/python3.7/site-packages/ctapipe_io_lst/__init__.py", line 259, in fill_lst_event_container_from_zfile
    unpacked_cdts =  struct.unpack(rec_fmt, event.lstcam.cdts_data)
struct.error: unpack requires a buffer of 31 bytes
```
This seems to be due to the recent upgrade of DAQ data model (extraction of CDTS data). Now it should be changed according to this structure:

```
typedef struct
{
    tUInt64 uctsTimeStamp;
    tUInt32 uctsAddress;
    tUInt32 eventCounter;
    tUInt32 busyCounter;
    tUInt32 ppsCounter;
    tUInt32 clockCounter;
    tUInt8 triggerType;                 // For TIB, this is the first 7 bits of SPI
    tUInt8 whiteRabbitResetBusyStatus;  // A few status flags here, see below
    tUInt8 stereoPattern;               // For TIB, this is the next 8 bits of the SPI string
    tUInt8 numInBunch;                  // This information only needed for debugging
    tUInt32 cdtsVersion;                // Should be x.y.z where x is 2bytes and y, z are a byte each
} __attribute__((__packed__)) tExtDevicesCDTSMessage;
```

morcuended:
Well...it seems that this only affects to runs from 20191206 because I could process runs from 20191209 without problems.

So false alarm. I will keep this open in case this is useful for anyone else.

morcuended:
It turns out that only run1720 is affected by this format change.

morcuended:
I open this issue again.

[...]

PR #19, opened 2019-12-13
New time and geometry

FrancaCassol:
This PR change the default geometry to the version 3 (in ctapipe_extra) which is in the Camera frame, version 2 was wrongly rotated
I also set the r0 trigger_time to the dragon time of the central module (to be changed when we get a correct ucts)


FrancaCassol:
The problem here is that the correct geometry (version 3) is for the moment only in the master of ctapipe_extra, I will ask to get a new release but I don't know how long it will take. Should I add this geometry somewhere else?

FrancaCassol:
WARNING: it seems that the CDTS data model changed in the last version of the EvB. I am going to add to this PR the necessary change.

morcuended:
See #18

morcuended:
Hi @FrancaCassol 
This is the new structure according to Julien:
```
typedef struct
{
    tUInt64 uctsTimeStamp;
    tUInt32 uctsAddress;
    tUInt32 eventCounter;
    tUInt32 busyCounter;
    tUInt32 ppsCounter;
    tUInt32 clockCounter;
    tUInt8 triggerType;                 // For TIB, this is the first 7 bits of SPI
    tUInt8 whiteRabbitResetBusyStatus;  // A few status flags here, see below
    tUInt8 stereoPattern;               // For TIB, this is the next 8 bits of the SPI string
    tUInt8 numInBunch;                  // This information only needed for debugging
    tUInt32 cdtsVersion;                // Should be x.y.z where x is 2bytes and y, z are a byte each
} __attribute__((__packed__)) tExtDevicesCDTSMessage;
```

FrancaCassol:
Hi @morcuended,
thanks, I had the luck to get it too. I hope in the future this type of changes manage to follow an official communication channel

morcuended:
It would be possible to make it backward compatible? For example, check the version of the CDTS first, and then use the corresponding data model

FrancaCassol:
obviously, I will use the daq version to correctly read the data

mdpunch:
> Hi @morcuended,
> thanks, I had the luck to get it too. I hope in the future this type of changes manage to follow an official communication channel

Hello Franca,
What should be the official communication channel for this information?  As I'm not in the LST consortium, I don't know.  Julien is in LST, though, but doesn't seem to be on github, so won't get this message unless you tell him in the CPPM lab!
Good Luck,
  Michael.

FrancaCassol:
Hi @mdpunch,

> What should be the official communication channel for this information? As I'm not in the LST consortium, I don't know. Julien is in LST, though, but doesn't seem to be on github, so won't get this message unless you tell him in the CPPM lab!

I think you got the point. Don't worry, I think the message arrived.

morcuended:
Is the test failing because the current file `example_LST_R1_10_evts.fits.fz` does not contain the new-named `LSTCam-003.camgeom` table? Should it be updated then or add another LST file for testing?

MaxNoe:
> Is the test failing because the current file example_LST_R1_10_evts.fits.fz does not contain the new-named LSTCam-003.camgeom table? 

The CameraGeometry is in ctapipe-extra and has nothing to do with the input data file.

I would suggest that you add the lst camera geometry as a resource file in this repository, as it is clearly required for the lst event source.

FrancaCassol:
> I would suggest that you add the lst camera geometry as a resource file in this repository, as it is clearly required for the lst event source.
> 
> (pinging @kosack here)

Hi @MaxNoe,
that is a good idea. 
@rlopezcoto what do you think about that?

MaxNoe:
It's only 13 kB, so basically nothing

rlopezcoto:
it sounds like a good solution.

> > I would suggest that you add the lst camera geometry as a resource file in this repository, as it is clearly required for the lst event source.
> > (pinging @kosack here)
> 
> Hi @MaxNoe,
> that is a good idea.
> @rlopezcoto what do you think about that?

MaxNoe:
See https://github.com/cta-observatory/ctapipe_io_lst/pull/20

mdpunch:
> Hi @mdpunch,
> 
> > What should be the official communication channel for this information? As I'm not in the LST consortium, I don't know. Julien is in LST, though, but doesn't seem to be on github, so won't get this message unless you tell him in the CPPM lab!
> 
> I think you got the point. Don't worry, I think the message arrived.

Hi Franca,
Okay, good, but I'd like to know also!  
I was thinking it should be put in some data model repository like the one that Etienne Lyard started (and read there rather than hardcoded), but can't find that now,
Good Luck,
  Michael.

FrancaCassol:
Hi @mdpunch, 

> Okay, good, but I'd like to know also!

this is a big problem in my view, I also could not find the official place where this information is supposed to be. I got the data model one day from Julien (that is my only known official channel) and I had to limit my ambition to at least be informed if the output of the EvB is changing (in order to be able to modify the offline reader). I agree that this is not really satisfactory.

PR #21, opened 2019-12-16
Add automatic version

rlopezcoto:
Add automatic version loading as in `ctapipe` and `cta-lstchain` with the `version.py` file

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,11 +7,11 @@ Install:
 ```
 conda env create -n lstenv -f py3.7_env.yaml
 conda activate lstenv
-pip install .
+pip install -e .
 ```
 
 Or in your favorite conda environment:
 ```
 conda env update -f py3.7_env.yaml
-pip install .
+pip install -e .
 ```
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -25,8 +25,10 @@ from ctapipe.core.traits import Int, Bool
 from .containers import LSTDataContainer
 from ctapipe.io.containers import PixelStatusContainer
 
-__all__ = ['LSTEventSource']
+from . import version
+__version__ = version.get_version(pep440=False)
 
+__all__ = ['LSTEventSource']
 
 def load_camera_geometry(version=3):
     ''' Load camera geometry from bundled resources of this repo '''
--- a/ctapipe_io_lst/version.py
+++ b/ctapipe_io_lst/version.py
@@ -1,0 +1,170 @@
+"""
+Get version identification from git.
+
+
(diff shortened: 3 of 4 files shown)


MaxNoe:
This is not good!

Importing a module in setup.py that directly imports dependencies makes it impossible for setuptools to install those dependencies. Running pip install will directly fail with an import error.

rlopezcoto:
Do you refer to the import of `ctapipe_io_lst`?
it's been working like this for `lstchain` up to now, but we can modify it, can you work out a suggestion that we can add here and in lstchain?
Thanks!

MaxNoe:
Probably because you installed all dependencies before by hand. But running `pip install .` now raises import errors for e.g. astropy if in a fresh environment.

The import makes it impossible for setuptools to install the dependencies listed in `install_requires` before importing the package:

```
$ pip install .
Processing /home/maxnoe/Uni/CTA/ctapipe_io_lst
    ERROR: Command errored out with exit status 1:
     command: /home/maxnoe/.local/anaconda3/envs/test/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-req-build-k1ljh_jn/setup.py'"'"'; __file__='"'"'/tmp/pip-req-build-k1ljh_jn/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-req-build-k1ljh_jn/pip-egg-info
         cwd: /tmp/pip-req-build-k1ljh_jn/
    Complete output (7 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-k1ljh_jn/setup.py", line 3, in <module>
        import ctapipe_io_lst
      File "/tmp/pip-req-build-k1ljh_jn/ctapipe_io_lst/__init__.py", line 7, in <module>
        import numpy as np
    ModuleNotFoundError: No module named 'numpy'
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

This should not happen!

rlopezcoto:
ok, I see, thanks again for reporting. It would be great if you could work out a solution, but if you cannot make it, I'll give it a try because we need a new release of the repo to read all new data. Are you working on something @MaxNoe ?

MaxNoe:
Yes, I'll get back to you.

MaxNoe:
See #22

PR #22, opened 2019-12-17
Fix pip install

MaxNoe:
* Add link to prozozfits reader (there is a newer version, do you want to still use 1.4.2?)
* Duplicate version.py so it can be used at runtime and at install time (a bit ugly, but works)
* Define OPTICS  instead of using `OpticsDescription.from_name`, so this works without ctapipe-extra

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .pytest_cache
+ctapipe_io_lst/_version_cache.py
 
 # Compiled files
 *.py[co]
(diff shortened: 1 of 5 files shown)


MaxNoe:
No idea, just saw it when looking for the lonk

MaxNoe:
https://github.com/cta-sst-1m/protozfitsreader/releases

rlopezcoto:
I think this can be merges @MaxNoe, thanks a lot for the fix! After this, we will make a new release

PR #32, opened 2020-04-27
Move to next ctapipe release

MaxNoe:
* Make subarray a property so abc is happy


.travis.yml
```diff
@@ -1,18 +1,11 @@
...
-      env:
-        - PYTHON_VERSION=3.6
-        - CTAPIPE_VERSION=v0.7.0
-
     - os: linux
       python: 3.7
       env:
         - PYTHON_VERSION=3.7
-        - CTAPIPE_VERSION=v0.7.0
+        - CTAPIPE_VERSION=master
```
rlopezcoto, line 8: why not leaving the test agains the latest release plus that of the master?

MaxNoe: Because the first change we will make to use the master will break compatiblity with 0.7, so why keep testing against that?

Or do you want to keep compatibility to both? That will be a lot of work and pretty ugly.

rlopezcoto: whenever ctapipe v0.8.0 is released, we can update the version of the latest release, but if we only leave the `master` in travis, tests will not pass whenever they do not become compatible

MaxNoe: E.g. ctapipe.io.containers → ctapipe.containers

MaxNoe: As soon as 0.8 is released, 0.8 and master should be tested

rlopezcoto: ok, I see your point, I would still like to keep a reminder that whenever this is merged into ctapipe_io_lst `master`, we should add the ctapipe `v0.8.0` test -> maybe comment it in?

MaxNoe: In fact, this should be done **before** this is merged, to show that this is really compatible with 0.8.

rlopezcoto: so what about if instead of deleting, we comment those lines?

MaxNoe: Ok, I simply added the test for ctapipe 0.8. It will fail until ctapipe 0.8 is released

MaxNoe:
Converting to a draft until the obs id discussion is settled.

rlopezcoto:
I would merge this and make a new ctapipe_lst_io release, otherwise we cannot move forward with the lstchain update. 
Not sure what is the `obs_id` discussion @MaxNoe was referring to (maybe @FrancaCassol can comment) and he may not be able to respond, but if nobody opposes, we'll move forward with this so next week we can work on the lstchain update.

moralejo:
> I would merge this and make a new ctapipe_lst_io release, otherwise we cannot move forward with the lstchain update.
> Not sure what is the `obs_id` discussion @MaxNoe was referring to (maybe @FrancaCassol can comment) and he may not be able to respond, but if nobody opposes, we'll move forward with this so next week we can work on the lstchain update.

The discussion was whether we can for now consider the run number as our "obs_id"
I see no objection for doing so.

MaxNoe:
Des, just replace  the return -1 with this configuration_id, like in the master

rlopezcoto:
Done, thanks @MaxNoe!

PR #33, opened 2020-05-28
set obs_id to camera_config.configuration_id

MaxNoe:
Follow up from https://github.com/cta-observatory/cta-lstchain/issues/424

### DIFF
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -399,7 +399,7 @@ class LSTEventSource(EventSource):
         """
         container = self.data.r0
 
-        container.obs_id = -1
+        container.obs_id = self.camera_config.configuration_id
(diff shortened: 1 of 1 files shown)


FrancaCassol:
Hi @MaxNoe,
I am not sure that the run_id (which for the moment is a LST number) is supposed to be the obs_id (which I think is much more general at CTA level). The run_id is already saved in the LST container. Is there a reason to set it also at the obs_id level in the reader?

MaxNoe:
The combination of obs_id and event_id must uniquely identify an event. 

And yes, for now the run number is the obs_id for LST.

Why would you think it's inappropriate to fill this here?

MaxNoe:
I think it's settled now, especially with the comments made by @kosack 

>  But I'd still copy the value to obs_id.

@FrancaCassol ok?

PR #35, opened 2020-07-06
Adapt to new instrument description hierarchy

moralejo:
First steps towards having up-to-date subarray-telescope-camera- description


moralejo:
Should be ready now

PR #37, opened 2020-07-06
clean env file

vuillaut:


rlopezcoto:
tests are passing, but Travis info is not reaching back the repo and we could not merge, I temporarily deactivate the “Require status checks to pass before merging” to move forward

PR #39, opened 2020-07-09
Fix subarray tel positions path

vuillaut:
This is required for the subarray writing to work with ctapipe v0.8  and the dl1 writer.


ctapipe_io_lst/__init__.py
```diff
@@ -201,7 +201,7 @@ def create_subarray(self, tel_id=1):
 
         subarray = SubarrayDescription("LST1 subarray")
         subarray.tel_descriptions = tel_descriptions
-        subarray.tel_positions = tel_positions
+        subarray.positions = tel_positions
         subarray.tel[tel_id] = lst_tel_descr
```
MaxNoe, line 205: Just instantiating the `SubarrayDescription` with all info directly is more efficient and would have directly shown this bug.

Issue #25, opened 2020-01-22
use ctapipe.core.provenance to track open files


kosack:
I notice that there are multiple files opened at various times in this code.  Any time you open a file, you must register it with ctapipe's provenance system, so that we can track what depends on what (a requirement of CTA).

So if you *don't* use `ctapipe.utils.get_dataset_*` (which do it automatically), you must do it manually as follows:

```py3
from ctapipe.core import Provenance


...
# for inputs:
Provenance().add_input_file("LSTCam.camgeom.fits", role="CameraGeometry") 

# for outputs:
Provenance().add_output_file("events.h5", role="DL1/Event") 
```

Where role shuold be something that describes why you opened the file (e.g. "DL0/Event" or "CameraGeometry", etc.)

Then it will end up automatically in the Tool output (in provenance.log and in stdout if you set the `--log-level=DEBUG`

The recorded provenance from `ctapipe-camdemo --camera MAGICCamMARS --log-level=DEBUG` looks like this:
```
...
      "input": [
         {
            "url": "/Users/kkosack/Library/Caches/Python-Eggs/ctapipe_extra-0.2.18-py3.7.egg-tmp/ctapipe_resources/MAGICCamMars.camgeom.fits.gz",
            "role": "dl0.tel.svc.camera"
         },
         {
            "url": "/Users/kkosack/Library/Caches/Python-Eggs/ctapipe_extra-0.2.18-py3.7.egg-tmp/ctapipe_resources/optics.ecsv.txt",
            "role": "dl0.tel.svc.optics"
         }
      ],
      "output": [],
...
```

PR #45, opened 2020-07-27
Add provenance tracking missing for input files

Bultako:
This PR fixes issue https://github.com/cta-observatory/ctapipe_io_lst/issues/25 adding missing provenance tracking for input files opened in several parts of the code. It also fixes the duplicated presence of a ``version.py`` file, and minor formatting aspects like import sorting or typos.

### DIFF
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -45,6 +45,7 @@ def load_camera_geometry(version=4):
     f = resource_filename(
         'ctapipe_io_lst', f'resources/LSTCam-{version:03d}.camgeom.fits.gz'
     )
+    Provenance().add_input_file(f, role="CameraGeometry")
     return CameraGeometry.from_table(f)
 
 
@@ -68,6 +69,7 @@ def read_pulse_shapes():
                                    'resources/no_corrections_pulse_LST.dat')
 
     data = np.genfromtxt(infilename, dtype='float', comments='#')
+    Provenance().add_input_file(infilename, role="PulseShapes")
     daq_time_per_sample = data[0, 0] * u.ns
     pulse_shape_time_step = data[0, 1] * u.ns
 


setup.py
```diff
@@ -1,6 +1,6 @@
 from setuptools import setup, find_packages
 from os import path
-from version import get_version, update_release_version
+from ctapipe_io_lst.version import get_version, update_release_version
```
MaxNoe, line 3: This breaks the installation in a fresh environment, please revert.

There was a reason, why the version code, unfortunately, had to be duplicated.

Bultako:
Thanks @MaxNoe 
I've reverted to the first commit.
I'll try to fix the duplicated `version.py` file issue in another PR

MaxNoe:
@Bultako There were some requirements that I couldn't fulfill with anything I tried. 

But I think, we should just use setuptools_scm, let's see if we can use that to tick all the boxes.
However, I fear that doing `pip install <url to github tar ball>` will always result in an unknown version unless we hardcode a fallback version somewhere.

Bultako:
@MaxNoe 
why not doing as others do?  i.e. `setup.py` file in `ctapipe`, which is adding the root dir to the path, this would result in

```
sys.path.insert(0, os.path.join(os.path.dirname(__file__)))
import ctapipe_io_lst # noqa

ctapipe_io_lst.version.update_release_version()
version = ctapipe_io_lst.version.get_version(pep440=True)
```

moreover, we are using the same `version.py` file, do we have different requirements?
if you think we can give it a try in this way I can open a new PR with this and continue discussion there.

MaxNoe:
Because ctapipe does not import any of its dependencies in ctapipe/__init__.py, but this package does. So importing it in the setup.py would break installation in a fresh environment.

What you could do, is add `dirname(__file__) / ctapipe_io_lst` to path and just import version.

Bultako:
Ok, let's continue with this in PR #47

rlopezcoto (approved):

PR #47, opened 2020-07-28
Remove duplicated version.py

Bultako:
This PR modifies code in `setup.py`, removes `version.py` file at the root dir and imports functions from the `version.py` file placed in `ctapipe_io_lst` folder instead.


rlopezcoto:
wouldn't this result in the same problem as in https://github.com/cta-observatory/cta-lstchain/issues/515?
maybe the solution pointed out by @MaxNoe: astropy/astropy#10774 is the best and we can implement it here and in `cta-lstchain`

Bultako:
@rlopezcoto 
The problem you cited above, about tagged releases not knowing its own version is already there.

```
$ wget https://raw.githubusercontent.com/cta-observatory/ctapipe_io_lst/v0.5.3/environment.yml
$ conda env create -n ctapipe_io_lst2 -f environment.yml
$ conda activate ctapipe_io_lst2
$ pip install https://github.com/cta-observatory/ctapipe_io_lst/archive/v0.5.3.zip
$ ipython

In [1]: import ctapipe_io_lst
In [2]: ctapipe_io_lst.__version__
Out[2]: 'unknown'
```

It is related with fetching the version with the `git` command https://github.com/cta-observatory/ctapipe_io_lst/blob/master/ctapipe_io_lst/version.py#L82

So, this PR will not add the already existing problem, though it will not resolve it neither... It will just refactor code and clean a duplicate scripts. I guess we could try resolving the issue with the solution you pointed in another PR.

rlopezcoto (approved):

maxnoe:
@rlopezcoto Actually, I would close this PR and open a new one that moves to using to `setuptools_scm` like we did recently with ctapipe, which also follows the approach taken by astropy.

rlopezcoto:
@maxnoe as it was explained by @Bultako this could work as a temporary removal of the duplicated `version.py` files, but if you prefer to go now for the permanent solution, it is also ok

maxnoe:
I just did it for ctapipe and is as easy to adapt as fixing the old, so I would skip that step.

maxnoe:
See #51

maxnoe:
This is obsolete now that #51 is merged, closing

PR #49, opened 2020-08-20
Replace struct loop with numpy.view and correct dtype

maxnoe:
This replaces the loop using struct to unpack a numpy array of bytes (uint8) with just using `view` to get the single arrays out of it.

In my test, this reduces the reading time for 10000 events from 20 to 16 seconds, so roughly a  20 % improvement.

The code is also much more readable and it preserves the correct data types (before it was converted to float64, which especially for the uint64 is not good!)

This is what I already mentioned here: #2

### DIFF
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -26,7 +26,6 @@ from ctapipe.containers import PixelStatusContainer
 from .containers import LSTDataContainer
 from .version import get_version
 
-from pkg_resources import resource_filename
 
 __version__ = get_version(pep440=False)
 __all__ = ['LSTEventSource']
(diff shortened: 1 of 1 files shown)


ctapipe_io_lst/__init__.py
```diff
@@ -40,6 +38,60 @@
...
+
+CDTS_BEFORE_37201_DTYPE = np.dtype([
+    ('event_counter', np.float32),
+    ('pps_counter', np.float32),
+    ('clock_counter', np.float32),
+    ('timestamp', np.float64),
+    ('camera_timestamp', np.float64),
+    ('trigger_type', np.uint8),
+    ('white_rabbit_status', np.uint8),
+    ('unknown', np.uint8),
```
MaxNoe, line 80: @FrancaCassol do you know what this 8th value here is? It was not used before

FrancaCassol: Hi @MaxNoe, looking back into the documentation, the name of this word was "arbitraryInformation".
In any case, it was never used.

FrancaCassol (approved):

PR #54, opened 2021-01-06
Move the storing of drs4 time shift in the apply_drs4_corrections function

FrancaCassol:
This is done in order to have the drs4 time shifts also for uncalibrated events (e.g during the processing of the flat-field events)


maxnoe:
This now always calculates the time correction for both gains. Before I used an optimized version for the case when gain selection happend that only computed the correction for the selected gain channel instead of both.

PR #56, opened 2021-01-28
Fix for lstchain#559

maxnoe:
This fixes the bug discussed here: https://github.com/cta-observatory/cta-lstchain/issues/559

Before and after comparison:

charge_comparison_now


rlopezcoto (approved):
please @jsitarek @pawel21 be aware that this has already been implemented here.

PR #57, opened 2021-01-29
Use tib trigger if available by default

maxnoe:
This is motivated by run 2006.0000, where after a UCTS trigger type
with value 42 the ucts trigger type seems to be shifted by one
without the UCTS time showing an offset.

Also adds  a warning if an event with UCTS trigger == 42 is encountered.

### DIFF
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -477,7 +477,19 @@ class LSTEventSource(EventSource):
         trigger.tels_with_trigger = [tel_id]
         trigger.tel[tel_id].time = trigger.time
 
-        trigger_bits = array_event.lst.tel[tel_id].evt.ucts_trigger_type
+        lst = array_event.lst.tel[tel_id]
+        tib_available = lst.evt.extdevices_presence & 1
+
+        # tib seems to be more reliable when available
+        if tib_available:
+            trigger_bits = lst.evt.tib_masked_trigger
+
+        else:
+            trigger_bits = lst.evt.ucts_trigger_type
+
+        if lst.evt.ucts_trigger_type == 42:
+            self.log.warning('Event with UCTS trigger_type 42 found. Probably means unreliable or shifted UCTS data.')
+
         # first bit mono trigger, second stereo.
         # If *only* those two are set, we assume it's a physics event
         # for all other we only check if the flag is present
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -224,6 +224,9 @@ class LSTEventSource(EventSource):
         self.time_calculator = EventTimeCalculator(subarray=self.subarray, parent=self)
         self.pointing_source = PointingSource(subarray=self.subarray, parent=self)
 
+        # by default, use ucts if available
+        self.ucts_42_found = False
+
     @property
     def subarray(self):
         return self._subarray
@@ -481,15 +484,21 @@ class LSTEventSource(EventSource):
         tib_available = lst.evt.extdevices_presence & 1
         ucts_available = lst.evt.extdevices_presence & 2
 
-        # tib seems to be more reliable when available
-        if tib_available:
+        # Since at least June 2020, ucts should be more reliable
+        # before there were runs when ucts was missing or was shifted
+        # by one event after an event with ucts_trigger_type == 42.
+        # In case that happens we swtich to tib triggers
+        if tib_available and self.ucts_42_found:
             trigger_bits = lst.evt.tib_masked_trigger
-        elif ucts_available:
+
+        elif ucts_available and not self.ucts_42_found:
             trigger_bits = lst.evt.ucts_trigger_type
             if lst.evt.ucts_trigger_type == 42:
+                self.ucts_42_found = True
                 self.log.warning(
                     'Event with UCTS trigger_type 42 found.'
                     ' Probably means unreliable or shifted UCTS data.'
+                    ' Switching to TIB if available.'
                 )
         else:
             self.log.warning('No trigger info available.')
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -169,7 +169,7 @@ class LSTEventSource(EventSource):
         help=(
             'Default source for trigger type information.'
             ' For older data, tib might be the better choice but for data newer'
-            ' than 2020-07, ucts is the preferred option. The source will still'
+            ' than 2020-06-25, ucts is the preferred option. The source will still'
             ' fallback to the other device if the chosen default device is not '
             ' available'
         )
@@ -515,7 +515,7 @@ class LSTEventSource(EventSource):
             self.log.warning(
                 'Event with UCTS trigger_type 42 found.'
                 ' Probably means unreliable or shifted UCTS data.'
-                ' Switching to TIB if available.'
+                ' Consider switching to TIB using `default_trigger_type="tib"`'
             )
 
         # first bit mono trigger, second stereo.
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -20,7 +20,7 @@ from enum import IntFlag, auto
 
 from ctapipe.io import EventSource
 from ctapipe.io.datalevels import DataLevel
-from ctapipe.core.traits import Int, Bool, List, Float
+from ctapipe.core.traits import Int, Bool, List, Float, Enum
 from ctapipe.containers import PixelStatusContainer, EventType
 
 from .containers import LSTArrayEventContainer, LSTServiceContainer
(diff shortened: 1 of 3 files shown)


moralejo:
Are you sure this is better in general, and not just for run 2006?  What type of run is 2006? It is not among those in the DL1 check, so is it not a sky run? In the following sky run, 2007, http://www.lst1.iac.es/datacheck/dl1/v0.6.1_v05/20200218/datacheck_dl1_LST-1.Run02007.pdf the DL1 check indeed shows faulty UCTS time and many 42 ucts event type tags, while TIB times and tags look more reasonable. 

But bear in mind that, at least for the data after June 25th (firmware fix related to jumping UCTS info) the UCTS tags, with the patch to solve jumps, are reliable - whereas I am not that sure about TIB. 

So if we want to always get the best possible source of event tags, I am afraid we have to go for something more complicated than just using TIB by default.

maxnoe:
This is the approach I discussed with @FrancaCassol, we were aware that tib might also not be reliable.

The other possibility we discussed was:

* Only switch to TIB by default if an event with `ucts_trigger_type == 42` is encountered

morcuended:
> What type of run is 2006? It is not among those in the DL1 check, so is it not a sky run?

Pedestal-calibration run.

From Isidro's data check this is the correlation between TIB and UCTS trigger types for that run (http://www.lst1.iac.es/datacheck/data/R1/RUNS_FROM_LST-1/SubRun_Plots/Run02006/Merged.Run02006.Subruns_from_0000_to_0004/Pairs_of_Trigger_Type_Distribution.Subruns_from_0_to_4.events_from_0_to_52999.pixels_from_0_to_1854.samples_from_2_to_37.png)

> The other possibility we discussed was:
> 
> Only switch to TIB by default if an event with ucts_trigger_type == 42 is encountered
> From what you write, this might be preferable? Or should we make it date (run?) dependent?

Not sure if we have some kind of statistics from all runs. I will ask Isidro. For data after June 25th, the first thing I can think of is using UCTS whenever availe, otherwise, stick with TIB triggers types.

mdpunch:
> Are you sure this is better in general, and not just for run 2006? What type of run is 2006? It is not among those in the DL1 check, so is it not a sky run? In the following sky run, 2007, http://www.lst1.iac.es/datacheck/dl1/v0.6.1_v05/20200218/datacheck_dl1_LST-1.Run02007.pdf the DL1 check indeed shows faulty UCTS time and many 42 ucts event type tags, while TIB times and tags look more reasonable.
> 
> But bear in mind that, at least for the data after June 25th (firmware fix related to jumping UCTS info) the UCTS tags, with the patch to solve jumps, are reliable - whereas I am not that sure about TIB.
> 
> So if we want to always get the best possible source of event tags, I am afraid we have to go for something more complicated than just using TIB by default.

My understanding from discussion with Luis Angel Tejedor at UCM is that this series of runs was after a new version of the TIB firmware, and that he found a bug (which gave rise to the "42"s, which are in fact a sign of SPI time-out).  As far as I know, this is fixed now.

I don't see Luis Angel on here, but it would be good to interact with him on this type of question,.

--- Extract of my mail to him on this below:

> In the data, there is the tm_time_valid bit - essentially that the PLL is locked and time synched to Master correctly, which I would call the White Rabbit Status.  In my decoding I have a "whiteRabbitResetBusyStatus", where I add the flag for rst_cnt_ack from the tailer of the bunch, which as described is set if the counters and TDC are enabled.  So I guess that the "White Rabbit Status" is in fact the "whiteRabbitResetBusyStatus", for which "3" means that all is okay.   In my decoding, I further add to the "whiteRabbitResetBusyStatus" if there is a busy flag (i.e. if the incoming trigger has come in through the "busy" input line), I add that as the MSB, and if there is a busy bit in the SPI, I put that as the next MSB.  I guess I should document this further in the description of the code (which is a bit outside the scope of the TiCkS user manual).
>
>So, the normal statuses (in decimal) are
>
>    "0b00000011" = "3" : WR locked okay, TDC & counters reset okay
>    "0b11000011" = "195" : as above, but event coming in on busy trigger line with busy bit set
>
>And you seem to be getting the abnormal status of
>
>    "0b01000011" = "67" : WR locked okay, TDC and counters reset okay, event on read-out trigger line, but busy bit set.
>
>My guess is that, as described in the description of the fields in the doc, if the SPI reception on the UCTS side times out, then the SPI is set to "0xAAAA", and this is probably giving the result you describe.   The bit in the status which changes the status to "67" is just because the SPI timed out, and so was set to "0xAAAA".

moralejo:
> 
> My understanding from discussion with Luis Angel Tejedor at UCM is that this series of runs was after a new version of the TIB firmware, and that he found a bug (which gave rise to the "42"s, which are in fact a sign of SPI time-out). As far as I know, this is fixed now.

I am not sure whether the fix Luis Ángel refers to is the one that was made in late June.  Following that fix, from July 2020 on, the UCTS info is correct, except for occasional jumps that we managed to solve via a software correction (already ported by @maxnoe to the LST R1 event source). It is just for older data that we have an issue. And it may well be that the TIB info is more reliable for those older data. But for the newer ones we do not have to change anything, it works as it is now.

maxnoe:
@moralejo I just changed to the following:

* Use ucts by default
* If an event with `ucts_trigger_type == 42` is encountered, switch to TIB.

moralejo:
> This is the approach I discussed with @FrancaCassol, we were aware that tib might also not be reliable.
> 
> The other possibility we discussed was:
> 
> * Only switch to TIB by default if an event with `ucts_trigger_type == 42` is encountered
> 
> From what you write, this might be preferable? Or should we make it date (run?) dependent?

I don't really know if occurrence of ucts_trigger_type 42 is good enough indicator.  My only point is we should only do that for pre-July 2020 data. Also for the data taken in November **2019** the UCTS trigger type, when available, seems reliable (judging from the charge distributions of pedestal-tagged events).

morcuended:
> My understanding from discussion with Luis Angel Tejedor at UCM is that this series of runs was after a new version of the TIB firmware, and that he found a bug (which gave rise to the "42"s, which are in fact a sign of SPI time-out). As far as I know, this is fixed now.

AFAIK, those 42-tagged events were appearing very rarely for some months now. However, after the recent TIB upgrade (mid-December 2020), the number of 42-tagged events taken with the new firmware was >50%. This should be fine after Luis Ángel's fix.

moralejo:
> > My understanding from discussion with Luis Angel Tejedor at UCM is that this series of runs was after a new version of the TIB firmware, and that he found a bug (which gave rise to the "42"s, which are in fact a sign of SPI time-out). As far as I know, this is fixed now.
> 
> AFAIK, those 42-tagged events were appearing very rarely for some months now. However, after the recent TIB upgrade (mid-December 2020), the number of 42-tagged events taken with the new firmware was >50%. This should be fine after Luis Ángel's fix.

The plot below, from the DL1 long-term check, shows the identification of pedestal events (from UCTS) going wrong after the night of Dec 15th ("12/16" in the graph) and recovering just before the last night that has been processed for this check (night of Dec 22nd). By looking at individual runs indeed one sees lots of 42 ucts trigger types in the bad week. I do not know if TIB would provide a good trigger type in those cases. Also, as you say, "42" also appears occasionally before, in runs in which ucts is labelling events properly, so perhaps just a few 42's should not be enough to dismiss UCTS.

pedestal_rates

moralejo:
I propose the following:

-  use as default the UCTS event type (with the already implemented fix for jumps), on the grounds that it is working for the data taken after June 25h (except for the ugly December week shown above)
- have a switch to use the TIB event type instead - this we can activate for older data when needed

I am a bit worried about automatic decision to use one or the other, not sure we can come up with a robust recipe for that.

maxnoe:
@moralejo this is implemented now.

ctapipe_io_lst/__init__.py
```diff
@@ -484,26 +492,32 @@ def fill_trigger_info(self, array_event, zfits_event):
         tib_available = lst.evt.extdevices_presence & 1
         ucts_available = lst.evt.extdevices_presence & 2
 
-        # Since at least June 2020, ucts should be more reliable
```
moralejo: Since 2020 June 25th

ctapipe_io_lst/__init__.py
```diff
@@ -484,26 +492,32 @@ def fill_trigger_info(self, array_event, zfits_event):
...
-        # before there were runs when ucts was missing or was shifted
-        # by one event after an event with ucts_trigger_type == 42.
```
moralejo: The one event jumps occur (not very frequently) also after June 25th, but are fixable

ctapipe_io_lst/__init__.py
```diff
@@ -484,26 +492,32 @@ def fill_trigger_info(self, array_event, zfits_event):
...
+
         else:
             self.log.warning('No trigger info available.')
             return
 
+        if ucts_available and lst.evt.ucts_trigger_type == 42:
+            self.log.warning(
+                'Event with UCTS trigger_type 42 found.'
+                ' Probably means unreliable or shifted UCTS data.'
+                ' Switching to TIB if available.'
```
moralejo, line 518: Not switching now. I would not even suggest to switch, since for post-June 25th data some 42's appear, yet the UCTS event types seem ok with the implemented fix.

moralejo (approved):
Just a couple of comments about code comments - for the rest it is fine for me.

ctapipe_io_lst/__init__.py
maxnoe: It is in the help text of the option

ctapipe_io_lst/__init__.py
maxnoe: right. I will add a comment about the option.

PR #59, opened 2021-02-11
Changes for interleaved events and old runs

FrancaCassol:
- Add trailet in order not to calibrate pedestals and FF events if requested
- Correct "if"  statement  for exception in case of missing ucts info, which is often the case in old runs

### DIFF
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -175,6 +175,11 @@ class LSTEventSource(EventSource):
         )
     ).tag(config=True)
 
+    calibrate_flatfields_and_pedestals = Bool(
+        default_value=True,
+        help='To be set to True for calibration processing'
+    ).tag(config=True)
+
     classes = [PointingSource, EventTimeCalculator, LSTR0Corrections]
 
     def __init__(self, input_url=None, **kwargs):
@@ -341,7 +346,13 @@ class LSTEventSource(EventSource):
 
             # gain select and calibrate to pe
             if self.r0_r1_calibrator.calibration_path is not None:
-                self.r0_r1_calibrator.calibrate(array_event)
+
+                # skip flatfield and pedestal events if asked
+                if (not (array_event.trigger.event_type == EventType.FLATFIELD or
+                    array_event.trigger.event_type == EventType.SKY_PEDESTAL) or
+                         self.calibrate_flatfields_and_pedestals):
+
+                    self.r0_r1_calibrator.calibrate(array_event)
 
             yield array_event
 
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -348,10 +348,10 @@ class LSTEventSource(EventSource):
             if self.r0_r1_calibrator.calibration_path is not None:
 
                 # skip flatfield and pedestal events if asked
-                if (not (array_event.trigger.event_type == EventType.FLATFIELD or
-                    array_event.trigger.event_type == EventType.SKY_PEDESTAL) or
-                         self.calibrate_flatfields_and_pedestals):
-
+                if (
+                    array_event.trigger.event_type not in {EventType.FLATFIELD, EventType.SKY_PEDESTAL}
+                        or self.calibrate_flatfields_and_pedestals
+                ):
                     self.r0_r1_calibrator.calibrate(array_event)
 
             yield array_event
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -6,6 +6,7 @@ import numpy as np
 from astropy import units as u
 from pkg_resources import resource_filename
 import os
+from astropy.time import Time
 from os import listdir
 from ctapipe.core import Provenance
 from ctapipe.instrument import (
@@ -178,6 +179,11 @@ class LSTEventSource(EventSource):
     calibrate_flatfields_and_pedestals = Bool(
         default_value=True,
         help='To be set to True for calibration processing'
+    ).tag(config=True)
+
+    fill_timestamp = Bool(
+        default_value=True,
+        help='To be set to False for data without ucts info'
     ).tag(config=True)
 
     classes = [PointingSource, EventTimeCalculator, LSTR0Corrections]
@@ -495,7 +501,10 @@ class LSTEventSource(EventSource):
         tel_id = self.tel_id
 
         trigger = array_event.trigger
-        trigger.time = self.time_calculator(tel_id, array_event)
+        if self.fill_timestamp:
+            trigger.time = self.time_calculator(tel_id, array_event)
+        else:
+            trigger.time = Time(0, format='mjd', scale='tai')
         trigger.tels_with_trigger = [tel_id]
         trigger.tel[tel_id].time = trigger.time
 
--- a/ctapipe_io_lst/calibration.py
+++ b/ctapipe_io_lst/calibration.py
@@ -13,7 +13,7 @@ from ctapipe.core.traits import (
 )
 
 from ctapipe.calib.camera.gainselection import ThresholdGainSelector
-from ctapipe.containers import MonitoringContainer, ArrayEventContainer
+from ctapipe.containers import MonitoringContainer, ArrayEventContainer, EventType
 from ctapipe.io import HDF5TableReader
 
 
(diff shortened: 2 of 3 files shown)


ctapipe_io_lst/calibration.py
```diff
@@ -136,6 +136,11 @@ class LSTR0Corrections(TelescopeComponent):
         help='Threshold for the ThresholdGainSelector.'
     ).tag(config=True)
 
+    calibrate_ff_and_ped = Bool(
+        default_value=True,
+        help=('To be set to True for calibration processing')
```
maxnoe, line 141: parentheses not needed here

ctapipe_io_lst/calibration.py
```diff
@@ -205,6 +210,13 @@ def update_first_capacitors(self, event: ArrayEventContainer):
             )
 
     def calibrate(self, event: ArrayEventContainer):
+
+        if ((event.trigger.event_type == EventType.FLATFIELD or
+             event.trigger.event_type == EventType.SKY_PEDESTAL) and
+            not self.calibrate_ff_and_ped):
```
maxnoe, line 216: I don't really like that you convert this method into a no-op depending on this options.

I think it would be better to move the traitlet to the source and then do:

```
if self.calibrate_ff_and_ped or event.trigger.event_type == EventType.SUBARRAY:
    self.lstr0corrections.calibrate(event)
```

maxnoe (changes requested):

ctapipe_io_lst/event_time.py
```diff
@@ -180,7 +180,7 @@ def __call__(self, tel_id, event):
 
         # first event and values not passed
         else:
-            if not ucts_available:
+            if not ucts_available and self.timestamp == 'ucts':
```
maxnoe, line 183: No! The code below uses ucts to create tib and dragon time. So you cannot only raise the exception when `self.timestamp == 'ucts'`.

This codepath needs to run for tib and dragon timestamps and simply does not work if ucts is not available.

ctapipe_io_lst/calibration.py
FrancaCassol: ok for me, but are we sure there are not other triggers we like to calibrate?

maxnoe: Sorry, the proposal was mainly to move it to the source, not in here. If you have a more appropriate check (like explictly checking for the pedestals or flatfield) go ahead.

ctapipe_io_lst/event_time.py
FrancaCassol: Ok @maxnoe , I did not dig inside all these methods, please suggest you a way to process files without ucts data, for the moment it is not possible as far as I understand (perhaps I am missing a way?)

maxnoe: We discussed it in slack: add an option to the source to not fill the timestamp.

FrancaCassol: That is doesn't work, because I need the timestamp and for those runs I have it from the tib

FrancaCassol: why I should not use it?

maxnoe: There is now way to get a correct timestamp without ucts.

ctapipe_io_lst/calibration.py
FrancaCassol: Ok, I move it there (I was indeed undecided)

ctapipe_io_lst/event_time.py
maxnoe: So the question is for what do you need timestamps?

FrancaCassol: > There is no way to get a correct timestamp without ucts. See discussion here:
> [cta-observatory/cta-lstchain#581](https://github.com/cta-observatory/cta-lstchain/issues/581)

They were the only one accessible for several months, so we decided to through everything away? 
Then, the problem is solved. I do not think it makes sense to analyse data for which we do not fill the timestamp. 

Only, we must decide a new time reference run, because the 1625 can not be used anymore. What do you think @moralejo ?

morcuended: > I do not think it makes sense to analyse data for which we do not fill the timestamp.

I agree. I usually do not do it and when I do it just for completeness I note it down in the analysis logbook we have.

Only, we must decide a new time reference run, because the 1625 can not be used anymore. What do you think @moralejo ?

I've been producing the time calibration files every day for some time now. When it is now possible I use 1625. Wanted to ask you about this back then, but never did. I hope it is not a problem.

FrancaCassol: In any case, it seems to me that a timestamp with global offset  is better than zero timestamp

maxnoe: I disagree. A wrong timestamp that looks like it could be right is worse than knowing there is no correct timestamp.

maxnoe: We could however, allow filling the invalid timestamp `0 MJD TAI`

FrancaCassol: Hi @morcuended,

> I've been producing the time calibration files every day for some time now. When it is not possible I use 1625. Wanted to ask you about this back then, but never did. I hope it is not a problem.

These corrections are supposed to be stable with time and change only of there is some hardware change. So  the idea was to keep one reference file  for each hardware period (and we started with 1625) in order to be sure that the time coefficients were properly estimated and the camera setup was ok. But it is true that at end we remained fixed to it. Did I understand correctly that you are now using the time coefficients of the calibration run of each night?

morcuended: > Did I understand correctly that you are now using the time coefficients of the calibration run of each night?

Correct

FrancaCassol: > I disagree. A wrong timestamp that looks like it could be right is worse than knowing there is no correct timestamp.

It is only an offset of some seconds (for these commissioning data it is not so critical) and if we ask expressively to use tib, we take a risk under control (it is not going to tib by default).

moralejo: I agree with @FrancaCassol, and we do not want to be unable to process some data with the new release.
If I understand properly, the example Franca mentioned is that of a calibration run. We will not need accurate timing for such a run (or even any timing at all, since for now we will just use the whole subrun's interleaved events to locate stars and modify cleaning levels). But definitely we want to be able to process it.

maxnoe: This is what I proposed in the first place, an option not to fill timing info (or fill clearly invalid), but having timestamps that is wrong by 8 to 120 seconds is something I would not do.

moralejo: But then, @FrancaCassol, why is the timestamp needed?

maxnoe: If it is about relative time, you can just directly access the counter values from  `event.lst.evt`.

But filling `trigger.time` with a wrong absolute timestamp is not a good idea.

FrancaCassol: They are used in the code: there is an option to select the calibration statistics  based on a time interval not on the number of the events. In this case a global shift does not hurt, because only relative times are considered.

FrancaCassol: Let's say, the important thing is to fill a timestamp. I would keep the tib value (knowing that I am doing it and the risks I take) but to de-block the situation a zero value will also work

ctapipe_io_lst/__init__.py
```diff
@@ -341,7 +346,13 @@ def _generator(self):
 
             # gain select and calibrate to pe
             if self.r0_r1_calibrator.calibration_path is not None:
-                self.r0_r1_calibrator.calibrate(array_event)
+
+                # skip flatfield and pedestal events if asked
+                if (not (array_event.trigger.event_type == EventType.FLATFIELD or
+                    array_event.trigger.event_type == EventType.SKY_PEDESTAL) or
+                         self.calibrate_flatfields_and_pedestals):
```
maxnoe, line 353: This might be more readable:

```
if array_event.trigger.event_type not in {EventType.FLATFIELD, EventType.SKY_PEDESTAL} or self.calibrate_flatfields_and_pedestals
```

FrancaCassol: :-) I knew you had a better way ;-)

maxnoe (approved):

PR #70, opened 2021-02-22
Make sure dtype of night summary table is int for counters

maxnoe:
Columns will always have integer dtype where appropriate and where the fill contains `nan` these are converted to masked array elements.

### DIFF
--- a/ctapipe_io_lst/event_time.py
+++ b/ctapipe_io_lst/event_time.py
@@ -1,13 +1,16 @@
+from collections import deque, defaultdict
+import numpy as np
+
+import astropy.version
+from astropy.io.ascii import convert_numpy
+from astropy.table import Table
+from astropy.table import Table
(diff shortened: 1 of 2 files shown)


ctapipe_io_lst/event_time.py
```diff
@@ -1,13 +1,16 @@
...
-from astropy.time import Time
-from astropy.table import Table
-import numpy as np
 from collections import deque, defaultdict
+import numpy as np
+
 import astropy.version
+from astropy.io.ascii import convert_numpy
+from astropy.table import Table
+from astropy.table import Table
```
FrancaCassol, line 7: two imports for Table

maxnoe: fixed, thanks

FrancaCassol (approved):

PR #72, opened 2021-02-22
Remove option to calculate timestamp from TIB, allow using svc.date as reference

maxnoe:
* Since times calculated based on the dragon counters are the most precise
  and it is always possible to calculate them, there is no reason to use
  TIB counters for time, so the option to calculate time based on TIB is
  removed.

* If reference counters / timestamps for dragon time are not given,
  the run start timestamp svc.date is used, which results in timestamps
  being off by an unknown amount of seconds. These timestamps can
  still be usefull for tracking relative times or interpolating
  pointing, but not for precise event time stamps.

### DIFF
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -27,7 +27,7 @@ from ctapipe.containers import PixelStatusContainer, EventType
 from .containers import LSTArrayEventContainer, LSTServiceContainer
 from .version import __version__
 from .calibration import LSTR0Corrections
-from .event_time import EventTimeCalculator
+from .event_time import EventTimeCalculator, INVALID_TIME
 from .pointing import PointingSource
 from .anyarray_dtypes import (
     CDTS_AFTER_37201_DTYPE,
@@ -504,7 +504,7 @@ class LSTEventSource(EventSource):
         if self.fill_timestamp:
             trigger.time = self.time_calculator(tel_id, array_event)
         else:
-            trigger.time = Time(0, format='mjd', scale='tai')
+            trigger.time = INVALID_TIME
         trigger.tels_with_trigger = [tel_id]
         trigger.tel[tel_id].time = trigger.time
 
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -181,11 +181,6 @@ class LSTEventSource(EventSource):
         help='To be set to True for calibration processing'
     ).tag(config=True)
 
-    fill_timestamp = Bool(
-        default_value=True,
-        help='To be set to False for data without ucts info'
-    ).tag(config=True)
-
     classes = [PointingSource, EventTimeCalculator, LSTR0Corrections]
 
     def __init__(self, input_url=None, **kwargs):
@@ -501,10 +496,7 @@ class LSTEventSource(EventSource):
         tel_id = self.tel_id
 
         trigger = array_event.trigger
-        if self.fill_timestamp:
-            trigger.time = self.time_calculator(tel_id, array_event)
-        else:
-            trigger.time = NAN_TIME
+        trigger.time = self.time_calculator(tel_id, array_event)
         trigger.tels_with_trigger = [tel_id]
         trigger.tel[tel_id].time = trigger.time
 
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -22,12 +22,12 @@ from enum import IntFlag, auto
 from ctapipe.io import EventSource
 from ctapipe.io.datalevels import DataLevel
 from ctapipe.core.traits import Int, Bool, Float, Enum
-from ctapipe.containers import PixelStatusContainer, EventType
+from ctapipe.containers import PixelStatusContainer, EventType, NAN_TIME
 
 from .containers import LSTArrayEventContainer, LSTServiceContainer
 from .version import __version__
 from .calibration import LSTR0Corrections
-from .event_time import EventTimeCalculator, INVALID_TIME
+from .event_time import EventTimeCalculator
 from .pointing import PointingSource
 from .anyarray_dtypes import (
     CDTS_AFTER_37201_DTYPE,
@@ -504,7 +504,7 @@ class LSTEventSource(EventSource):
         if self.fill_timestamp:
             trigger.time = self.time_calculator(tel_id, array_event)
         else:
-            trigger.time = INVALID_TIME
+            trigger.time = NAN_TIME
         trigger.tels_with_trigger = [tel_id]
         trigger.tel[tel_id].time = trigger.time
 
--- a/ctapipe_io_lst/event_time.py
+++ b/ctapipe_io_lst/event_time.py
@@ -22,6 +22,7 @@ if astropy.version.major == 4 and astropy.version.minor <= 2 and astropy.version
 
 
 CENTRAL_MODULE = 132
+INVALID_TIME = Time(0, format='mjd', scale='tai')
 
 
 # fix for https://github.com/ipython/traitlets/issues/637
@@ -141,6 +142,7 @@ class EventTimeCalculator(TelescopeComponent):
         default_value=None,
         help='UCTS timestamp of a valid ucts/tib counter combination'
     ).tag(config=True)
+
     tib_counter0 = TelescopeParameter(
         Int(allow_none=True),
         default_value=None,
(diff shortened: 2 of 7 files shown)


ctapipe_io_lst/event_time.py
```diff
@@ -22,6 +22,7 @@
 
 
 CENTRAL_MODULE = 132
+INVALID_TIME = Time(0, format='mjd', scale='tai')
```
FrancaCassol, line 25: Perhaps, better a more explicit  "ZER0_TIME"

ctapipe_io_lst/event_time.py
```diff
@@ -141,6 +142,7 @@ class EventTimeCalculator(TelescopeComponent):
         default_value=None,
         help='UCTS timestamp of a valid ucts/tib counter combination'
     ).tag(config=True)
+
```
FrancaCassol, line 145: I would add an help to the trailet 
`timestamp = TelescopeParameter(
        trait=Enum(['ucts', 'dragon', 'tib']), default_value='dragon'
    ).tag(config=True)`

the name is not enough self-explaining

FrancaCassol: In general, it would be better to add to the doc of the class a short description of the logics that it contains

ctapipe_io_lst/__init__.py
```diff
@@ -504,7 +504,7 @@ def fill_trigger_info(self, array_event):
         if self.fill_timestamp:
             trigger.time = self.time_calculator(tel_id, array_event)
         else:
-            trigger.time = Time(0, format='mjd', scale='tai')
+            trigger.time = INVALID_TIME
```
FrancaCassol, line 507: is this still necessary?

ctapipe_io_lst/event_time.py
maxnoe: Zero is just a detail, invalid is the meaning. So I prefer invalid because it adds semantics.

Someting like this:

```
table['time'] == INVALID_TIME
```

Is more clear on the meaning than

```
table['time'] == ZERO_TIME
```

won't you agree?

moralejo: > table['time'] == ZERO_TIME
> ```
> 
> won't you agree?

Indeed it seems much more clear to me.

FrancaCassol: if you like, I prefer to know, how it is set an invalid time, but it is a question of taste

maxnoe: I just realized we also have it in ctapipe, so we should just import instead of define our own.

moralejo:
> * If reference counters / timestamps for dragon time are given,
>   the run start timestamp svc.date is used, 

"are *not* given", I assume

maxnoe:
yes, indeed.

ctapipe_io_lst/event_time.py
maxnoe: both done

ctapipe_io_lst/tests/test_stage1.py
```diff
@@ -39,10 +41,8 @@ def test_stage1(tmpdir):
                 'drive_report_path': str(test_drive_report)
             },
             'EventTimeCalculator': {
-                'ucts_t0_dragon': int(run_info['ucts_t0_dragon']),
-                'dragon_counter0': int(run_info['dragon_counter0']),
-                'ucts_t0_tib': int(run_info['ucts_t0_tib']),
-                'tib_counter0': int(run_info['tib_counter0']),
+                'dragon_reference_time': int(run_info['ucts_t0_dragon']),
```
FrancaCassol, line 44: Do we still need the int() here?

maxnoe: no

maxnoe: sorry, yes, we do. Because it is a `numpy.int64` not a python `int`, and the `json` module cannot handle `numpy.int64`. So there is a different reason here for the int call then the reason in lstchain.

maxnoe: Without it, you get 'TypeError: Object of type int64 is not JSON serializable'

FrancaCassol: I see, and I guess you can not directly convert to a python int (which is the difference between the two?
) ...

maxnoe: I don't understand what you mean by this question, since doing `int(x)` *is the direct way of converting something to a python int*.

The difference is that the pure python `int` is arbitrary length, `np.int64` as the name suggests is restricted to 64 bits.

maxnoe: E.g. you can do this with a python int:

```
>>> 2**65
36893488147419103232
```

But doing the same with a numpy int will give you an overflow:
```
>>> np.int64(2)**65
0
```

FrancaCassol: > I don't understand what you mean by this question, since doing `int(x)` _is the direct way of converting something to a python int_.

I mean in the converter function, so that you do not have this further conversion here

maxnoe: No, because there we have an astropy table is basically a collection of numpy arrays as columns, so is always using numpy dtypes.

FrancaCassol:
The code is fine, but personally  I would have not overwritten the original lst values, like in 
[https://github.com/cta-observatory/ctapipe_io_lst/blob/aa6b6a7912f5d64e3fa8efbffbab14710a1ac680/ctapipe_io_lst/event_time.py#L354](https://github.com/cta-observatory/ctapipe_io_lst/blob/aa6b6a7912f5d64e3fa8efbffbab14710a1ac680/ctapipe_io_lst/event_time.py#L354) 
or [https://github.com/cta-observatory/ctapipe_io_lst/blob/aa6b6a7912f5d64e3fa8efbffbab14710a1ac680/ctapipe_io_lst/event_time.py#L359](https://github.com/cta-observatory/ctapipe_io_lst/blob/aa6b6a7912f5d64e3fa8efbffbab14710a1ac680/ctapipe_io_lst/event_time.py#L359)
and just correctly set the final R1 trigger and timestamp,  so as to keep access to the original raw data and to easily trace the (solved) failing cases

ctapipe_io_lst/tests/test_stage1.py
FrancaCassol: ok

maxnoe:
But at the later stage you don't have any means to detect ucts jumps. So I think it is better to fill in the corrected values.

FrancaCassol:
As said, it is not blocking, just a question of taste, I lake the idea to keep the possibility to access with the reader to the original R0 data.

FrancaCassol (approved):

maxnoe:
I don't really know what's better here. But when investigating such issues, you probably anyway need to look  into the raw protozfits data using the `protozfits.File`.

FrancaCassol:
> But when investigating such issues, you probably anyway need to look into the raw protozfits data using the `protozfits.File`.

This is exactly what I would like to avoid.

moralejo:
> But at the later stage you don't have any means to detect ucts jumps. So I think it is better to fill in the corrected values.

I agree. As for the trigger type, the ucts=>tib replacement @FrancaCassol mentioned above is only used for the rare events right at the ucts jumps. In previous lstchain they was just left as "cosmics" (a reasonable guess). We could as well set an error value. But in any case, it is so few events that it is completely negligible, I am fine with filling in tib in those cases.

For the timestamp, I am all for setting the dragon one, which would exactly be the ucts one, had it not failed!

So I think we can move on.

maxnoe:
> This is exactly what I would like to avoid.

Any reason in particular?

moralejo (commented):
Fine for me!
(we'll have to modify some things in lstchain because of the new names like dragon_reference_time etc)

maxnoe:
Yes, I'll open a PR

FrancaCassol:
> Any reason in particular?

Because it is better to have only one way to access the R0 data. 
What the reader does is to produce the R1 data, it does not need to modify the R0 data.

maxnoe:
The way to access the r0 data is protozfits

maxnoe:
> (we'll have to modify some things in lstchain because of the new names like dragon_reference_time etc)

https://github.com/cta-observatory/cta-lstchain/pull/616

moralejo:
> > (we'll have to modify some things in lstchain because of the new names like dragon_reference_time etc)
> 
> [cta-observatory/cta-lstchain#616](https://github.com/cta-observatory/cta-lstchain/pull/616)

I'll take care of the script tests and the json, that I was recently modifying.

moralejo:
> > > (we'll have to modify some things in lstchain because of the new names like dragon_reference_time etc)
> > 
> > 
> > [cta-observatory/cta-lstchain#616](https://github.com/cta-observatory/cta-lstchain/pull/616)
> 
> I'll take care of the script tests and the standard config json, that I was recently modifying.

Ok, I had not checked, you were faster. Thanks!

PR #74, opened 2021-03-01
Run flat field tagging on all event types

maxnoe:


ctapipe_io_lst/__init__.py
```diff
@@ -556,11 +556,9 @@ def fill_trigger_info(self, array_event):
     def tag_flatfield_events(self, array_event):
         # currently, tagging of flat field events does not work,
         # they are reported as physics events, here a heuristic identifies
-        # those events.
+        # those events. Since trigger types might be wrong due to ucts errors,
+        # we try to identify flat field events in all trigger types.
         # this does only work if data is calibrated
```
rlopezcoto, line 561: remove this comment as well?

rlopezcoto (approved):

ctapipe_io_lst/__init__.py
moralejo: The comment is ok, but like 561 is misleading: the filter works on drs4-calibrated waveform, not yet converted to p.e.

rlopezcoto: I meant `l.561`

maxnoe: But it's true. We look at R1 data, so the drs4 calibration step has to be applied before. What's the problem with the comment?

maxnoe: We could try it also on r0 if r1 is not there?

rlopezcoto: I understand "calibrated" the one that has already been converted to p.e. (low and high level calibration applied), not only low level drs4 corrections have been applied.

maxnoe: changed.

PR #76, opened 2021-03-04
Update protozfitsreader to 2.0

maxnoe:
I have just released the new protozfitsreader to pypi in version 2.0. It should be compatible without changes to ctapipe_io_lst


setup.py
```diff
@@ -7,7 +7,7 @@
     install_requires=[
         'astropy~=4.2',
         'ctapipe~=0.10.0',
```
rlopezcoto, line 9: what about `ctapipe>=0.10.4` ?

rlopezcoto (commented):
what are the main changes from `v1.5.0` to `v2.0`?
If there could be some unexpected problems due to apparently unrelated changes, I would prefer to keep `v1.5.0` for lstchain (and therefore keep it compatible with ctapipe_io_lst `v0.7.0`)

maxnoe:
@rlopezcoto The major change is that the python bindings are no created via pybind11 and we correctly build manylinux wheels for pypi. So the absurd situation that the package shipped prebuild so files for each supported python version and architecture is no longer needed.

There should be no compatibility issues. We could have called it 1.6. But since the underlying binding technology and the repo location changed (now in the CTA gitlab) we called it 2.0.

I would highly recommend switching already for the next lstchain release.

In fact, this will allow us to also upload ctapipe_io_lst to pypi.

setup.py
maxnoe: ~= means >= 0.10.0,<0.11, so any 0.10.x version is supported. I think this is what we want. No need to require >= 0.10.4.

rlopezcoto:
> I would highly recommend switching already for the next lstchain release.
> 
> In fact, this will allow us to also upload ctapipe_io_lst to pypi.

ok, this is actually a very good reason to move to it. In any case, one can implement it, release a new `ctapipe_io_lst` version and if it is creating any conflicts with the tests, then we can reconsider to use this new `ctapipe_io_lst` version in the new release of lstchain

moralejo (approved):

maxnoe:
I think this is very unlikely, since the protozfitsreader tests are passing without changes, the tests here are passing without changes and I checked that the dl1 output of stage1 is the same.  

But you never have a guarantee that everything just works, so...

maxnoe:
Ah, and it also now supports python 3.7, 3.8 and 3.9 instead of 3.5 - 3.7.

PR #77, opened 2021-03-04
Add github CI config to deploy to PyPI

maxnoe:


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      # make sure we have version info
+      - run: git fetch --tags
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
```
rlopezcoto, line 20: why 3.8? in lstchain we are still requiring 3.7

rlopezcoto (commented):
Let's also follow the same you guys did in protozfits and make a new major release (`v0.8`) for the upload to pypi

.github/workflows/deploy.yml
maxnoe: Doesn't matter at all, since all it does is creating the tar balls.

maxnoe: >  in lstchain we are still requiring 3.7

Because protozfits only worked with 3.7, now you could change that ;)

maxnoe:
@rlopezcoto fine with me

rlopezcoto (approved):

PR #81, opened 2021-03-05
Fix dtypes of cdts data, fixes #80

### LINKED ISSUE #80: CDTS info has wrong dtypes
```
    ('address', np.float32), 
    ('event_counter', np.float32),
    ('busy_counter', np.float32),
    ('pps_counter', np.float32),

```

These should clearly be ints and the values are nonsensical right now due to this.

@FrancaCassol do you know where to get the correct dtypes from?

maxnoe:


rlopezcoto (approved):

maxnoe:
The dtype for the older version might be wrong. Need @FrancaCassol to double check and confirm

maxnoe:
Ok, Franca confirmed that this is indeed correct now.

PR #83, opened 2021-03-08
Force protozfits >2.0.1

rlopezcoto:
Thanks to @maxnoe and the `protozfits` for realising a very quick fix for https://github.com/cta-observatory/cta-lstchain/issues/620

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -18,4 +18,4 @@ dependencies:
   - h5py
   - pip:
       - pytest_runner
-      - protozfits~=2.0
+      - protozfits~=2.0.1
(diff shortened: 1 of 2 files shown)


maxnoe:
Currently the conda build is broken due to an oversight by me, it won't find the dependencies. Will be fixed in `.post` for this release and in the mean time you can do:

```
CMAKE_PREFIX_PATH=/path/to/conda/env pip install protozfits
```

maxnoe (approved):

maxnoe:
`.post1` is also now on pypi with the fix for conda builds

rlopezcoto:
> `.post1` is also now on pypi with the fix for conda builds

shall we include it explicitly in the environment? I can also make a new minor release to make sure we get the correct `protozfits` version

PR #84, opened 2021-03-09
Force protozfits >2.0.1.post1

rlopezcoto:
Due to https://github.com/cta-observatory/ctapipe_io_lst/pull/83#issuecomment-792785796

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ setup(
     install_requires=[
         'astropy~=4.2',
         'ctapipe~=0.10.0',
-        'protozfits~=2.0.1',
+        'protozfits~=2.0.1.post1',
         'setuptools_scm',
     ],
     package_data={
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -7,7 +7,7 @@
     install_requires=[
         'astropy~=4.2',
         'ctapipe~=0.10.0',
-        'protozfits~=2.0.1',
+        'protozfits~=2.0.1.post1',
```
maxnoe, line 10: No, this does not work. No you fix protozfits to `2.0.1` since the ~= operator only allows variation in the least significant version number you have included.

Since this release only fixes a build issue under specific cirmcumstances, I wouldn't even update the requirement.

Anyone installing it now will get the newest version anyway.

rlopezcoto: ok, sorry, I did not know, then I'll make the new release without this and include it in lstchain

PR #85, opened 2021-03-09
Expose service container at source

maxnoe:
Since the service container only depends on the `CameraConfiguration`, we can already fill it in `__init__` and expose it as a member of the source.

This will make it easier to deal with, for example like here in the PR for the pedestal fits writer:
https://github.com/cta-observatory/cta-lstchain/pull/601/


FrancaCassol:
Good idea

FrancaCassol (approved):

PR #90, opened 2021-03-13
Add protozfits from conda forge to environment.yml

maxnoe:


environment.yml
```diff
@@ -6,6 +6,7 @@ dependencies:
   - conda-forge::ctapipe=0.10
   - conda-forge::eventio
   - conda-forge::corsikaio
+  - conda-forge::protozfits=2.0
```
rlopezcoto, line 9: shouldn't it be `2.0.1`?

maxnoe: Again, I don't really want to nail the bug fix release here. If a newer but compatible version of protozfits is released, people will get it with this env without us having to change it.

rlopezcoto (approved):

Issue #91, opened 2021-03-25
Negative values in R0 data frame during the pedestal calculation


SeiyaNozaki:
It seems a few capacitors (2-3cell/(4096cel * 1855pix * 2gain)) have too small pedestal values (<50ADC).
For those capacitors, R0 data would be possibly negative during the pedestal calculation with time-lapse correction enabled.
(It is possible that raw data of those capacitors have already zero...)
The data type of R0 is uint16, so those negative values are converted to too high ADC counts, which will be saved in a pedestal table...

I'd like to avoid such a situation... possible solutions are:
1) convert the data type of R0 waveform into int16 (or float?) during pedestal computation
2) just fill zero if R0 data after time-lapse correction is negative

Do you have any opinions?

Related redmine ticket is:
https://forge.in2p3.fr/issues/44184

moralejo:
I don't understand what is the original content of those cells. Do they have baselines inconsistent with those obtained from DRS4 calibration?

In any case, we should absolutely avoid that a negative value becomes a large one. Set it to zero instead.
Do you think it is relevant to fix this before next reprocessing, or is it so rare that it is not urgent?

SeiyaNozaki:
Sorry, let me clarify:

Concerning the previous post, pedestal data (r3889) was taken without baseline correction.
So, the pedestal table which is currently used in EVB was created with this pedestal data. There is no extra offset (400), so a few capacitors would be zero at the raw level (, but I never noticed it...!). Then, those negative values after offline time-lapse correction were converted to high values, which increased the pedestal value of those capacitors for EVB baseline correction.

Then, we saw extremely high ADC counts(~65550) in R0 data with EVB baseline correction based on this pedestal table, which made three pixels outliers in the calibration script.  After EVB's baseline correction, those capacitors had negative value due to high pedestal and also converted to extremely high values already at the EVB stage.

I produced the pedestal table with the previous version of lstchain, and this table should work in EVB.
We try to take pedestal data with this pedestal file soon. If it works, the data of those capacitors will be correctly subtracted and extra 400 counts were added in EVB. So, there should be also no problem after this stage.

So, this is not an urgent issue, but we should fix it to produce a pedestal table for EVB in the future.

moralejo:
Ok, thanks for the clarification! In any case, let's implement a fix straightaway, to avoid that a negative value becomes a large positive one. Can you take care of it @SeiyaNozaki ?

SeiyaNozaki:
[...]

PR #92, opened 2021-03-26
change data type of R0 to int16 instead of uint16

SeiyaNozaki:
This PR is from the issue #91.
In this PR, I just changed the data type of R0 to ‘int16’ instead of ‘uint16’.
This has two merits:
1) When computing pedestal values without EVB baseline correction, R0 data is possibly negative after offline time-lapse correction. The current code would convert this negative one into too high values in uint16.

r3889

2) In EVB, the data type of R0 is defined as below:
> value = uint16((int16)sample + ((int16)offset - (int16)baseline))

So if the value is negative after EVB correction due to too high pedestal counts by accident or too low signal (<400ADC, e.g. undershoot of the pulse), too high counts are saved in the disk.
The new codes convert those high counts in uint16 to a corresponding negative value in int16.
For now, those too high counts are still kept in R1 stage since the data type of R1 is float32. 
r4224

### DIFF
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -623,7 +623,7 @@ class LSTEventSource(EventSource):
 
         # re-order the waveform following the expected_pixels_id values
         #  could also just do waveform = reshaped_waveform[np.argsort(expected_ids)]
-        dtype = reshaped_waveform.dtype
+        dtype = np.int16
         fill = np.iinfo(dtype).max
         reordered_waveform = np.full((self.n_gains, N_PIXELS, N_SAMPLES), fill, dtype=dtype)
         reordered_waveform[:, self.camera_config.expected_pixels_id, :] = reshaped_waveform


moralejo:
I do not understand one thing: why would one apply time lapse correction without (or before) baseline correction? Isn't that fundamentally wrong? If EVB does not apply the correction, the first thing one does in the offline analysis is to apply it. And then the fluctuation shown above would be safely above 0.

moralejo:
Also: do you understand why tests are not passing?

SeiyaNozaki:
>I do not understand one thing: why would one apply time lapse correction without (or before) baseline correction? Isn't that fundamentally wrong? If EVB does not apply the correction, the first thing one does in the offline analysis is to apply it. And then the fluctuation shown above would be safely above 0.

Here, I assume one would like to generate the pedestal table.
To find the baseline value, only time lapse correction is done and R0 container (not R1) is modified.
Around here:
https://github.com/cta-observatory/ctapipe_io_lst/blob/67b609ddcd8df495b6a2ac3f7e7fe2801f118106/ctapipe_io_lst/calibration.py#L427

For the normal offline correction, baseline/time lapse/spike corrections are done, and saved in R1 container (float32).

> Also: do you understand why tests are not passing?
No idea... What should I do...?

FrancaCassol:
Dear Seiya, 
your solution is ok for now. However, we should remember that  in a (hopefully) short future ACADA will like to receive from EvB a R1 waveform correctly fitting to   uint16 using the scale factors in eq 2.1 of https://edms.cern.ch/ui/file/2450834/2/CTA-SPE-COM-000000-0002_1e_R1-Event_Data_Model.pdf

moralejo:
> 
> Here, I assume one would like to generate the pedestal table.
> To find the baseline value, only time lapse correction is done and R0 container (not R1) is modified.
> Around here:
> https://github.com/cta-observatory/ctapipe_io_lst/blob/67b609ddcd8df495b6a2ac3f7e7fe2801f118106/ctapipe_io_lst/calibration.py#L427
>

Ah, ok, the "event-wise variable part" of the correction, which is the dt one, is taken out first to get a stable baseline per cell, which is then averaged to obtain the value which goes into the pedestal table.
 
> For the normal offline correction, baseline/time lapse/spike corrections are done, and saved in R1 container (float32).

ok

> 
> > Also: do you understand why tests are not passing?
> 
> No idea... What should I do...?

I'll have a look

moralejo:
I think it is this, in the tests:
https://github.com/cta-observatory/ctapipe_io_lst/blob/67b609ddcd8df495b6a2ac3f7e7fe2801f118106/ctapipe_io_lst/tests/test_lsteventsource.py#L98

moralejo:
But I am not sure if this change in the R0 waveform type may have any other consequences downstream, which perhaps are not covered by tests...

moralejo:
A more conservative change would be just to set the negative values to 0. If the values as rare (as you told me) the effect of this on the baseline calculation will be negligible anyway.

SeiyaNozaki:
Dear Franca, Abelardo,

Thanks for the comment!
>A more conservative change would be just to keep the uint16 type, but set the negative values to 0. If the negative values are rare (as you told me) the effect of this on the baseline calculation will be negligible anyway.

ok, then let me modify codes like this (set zero if it is negative) and keep uint16.

SeiyaNozaki:
I made a new PR (#93) with a new branch name just to avoid the confusion

maxnoe:
This is not the correct solution for this problem as it will remove all values from the waveforms larger than 2¹⁵, which are valid values in general! This is not ok and might bite us hard later.

If there is a specific issue with the time lapse calibration, it needs to be fixed there, not by butchering the waveform dtype.

PR #93, opened 2021-03-26
Set zeros if R0 data is negative after the time lapse correction

SeiyaNozaki:
continued from #92 
I changed a branch name just to avoid confusion.

### DIFF
--- a/ctapipe_io_lst/calibration.py
+++ b/ctapipe_io_lst/calibration.py
@@ -446,6 +446,12 @@ class LSTR0Corrections(TelescopeComponent):
             lst.svc.pixel_ids,
         )
 
+        # Set zeros if data is negative after time lapse correction to create a pedestal table
+        if waveform.dtype == np.uint16:
+            waveform_int16 = waveform.astype(np.int16)
+            waveform_int16[waveform_int16<0] = 0
+            waveform = waveform_int16.astype(np.uint16)
+
         container.waveform = waveform
 
     def interpolate_spikes(self, event, tel_id):


moralejo (approved):
Seems ok to me - and I assume you tested it in the case that raised the problem.

SeiyaNozaki:
Yes, I tested it and it seems working well (low pedestal value as expected for the capacitors which had a problem).

ctapipe_io_lst/calibration.py
```diff
@@ -446,6 +446,12 @@ def time_lapse_corr(self, event, tel_id):
             lst.svc.pixel_ids,
         )
 
+        # Set zeros if data is negative after time lapse correction to create a pedestal table
```
maxnoe, line 449: If the timelapse correction is the code introducing this problem, why not fix it in the time lapse calibration?

maxnoe:
This fix is not OK!

It reduces the dynamic range of the waveform from uint16 to int16 in general to fix a specific bug.

I will revert this.

maxnoe:
What is actually the problem here? If it is only the timelapse calib resulting in an underflow, it needs to be fixed there, not by removing all values > 2¹⁶ from the waveforms.

In general these are valid values!

maxnoe:
Also, there is now test here for this error condition!

moralejo:
> What is actually the problem here? If it is only the timelapse calib resulting in an underflow, it needs to be fixed there, not by removing all values > 2¹⁵ from the waveforms.
> 
> In general these are valid values!

As we two discussed in Slack: currently this poses no problem because there won't be genuine values above 2e15. But it is also true that eventually EVB might produce up-scaled values to fully use the available 16 bits. And this fix would then be problematic. So fixing it in the time lapse correction is a better solution.

PR #95, opened 2021-03-30
Fix underflow in timelapse calibration

maxnoe:
Proper fix to the time lapse calibration integer underflow issue reported here #93

### DIFF
--- a/ctapipe_io_lst/calibration.py
+++ b/ctapipe_io_lst/calibration.py
@@ -631,7 +631,9 @@ def do_time_lapse_corr(
 
                         # FIXME: Why only for values < 100 ms, negligible otherwise?
                         if time_diff_ms < 100:
-                            waveform[gain, pixel_id, sample] -= ped_time(time_diff_ms)
+                            # prevent underflow of the unsigned int value
+                            correction = min(ped_time(time_diff_ms), waveform[gain, pixel_id, sample])
+                            waveform[gain, pixel_id, sample] -= correction
 
                     # update the last read time
                     last_readout_time[gain, pixel_id, capacitor] = time_now
@@ -692,7 +694,9 @@ def do_time_lapse_corr_data_from_20181010_to_20191104(
                         time_diff_ms = time_diff / CLOCK_FREQUENCY_KHZ
 
                         if time_diff_ms < 100:
-                            waveform[gain, pixel_id, sample] -= ped_time(time_diff_ms)
+                            # prevent underflow of the unsigned int value
+                            correction = min(ped_time(time_diff_ms), waveform[gain, pixel_id, sample])
+                            waveform[gain, pixel_id, sample] -= correction
 
                 for sample in range(-1, N_SAMPLES - 1):
                     capacitor = (first_capacitor + sample) % N_CAPACITORS_PIXEL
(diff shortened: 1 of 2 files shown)


rlopezcoto (approved):

SeiyaNozaki:
Hi @maxnoe , thank you for pointing out my bad fix:)
The initial problem appeared when we produce a pedestal table for EVB, at that time the data type of `waveform` is uint16.
But, when we usually process the data from R0 to DL1, the data type of `waveform` is float for now and time-lapse correction is applied after pedestal subtraction. 
This PR prevents negative value for all the cases, no?

moralejo:
> Hi @maxnoe , thank you for pointing out my bad fix:)
> The initial problem appeared when we produce a pedestal table for EVB, at that time the data type of `waveform` is uint16.
> But, when we usually process the data from R0 to DL1, the data type of `waveform` is float for now and time-lapse correction is applied after pedestal subtraction.
> This PR prevents negative value for all the cases, no?

You mean the type of waveform should be checked, and this clipping of the correction should be applied only in case it is an int, right?

maxnoe:
> But, when we usually process the data from R0 to DL1, the data type of waveform is float for now and time-lapse correction is applied after pedestal subtraction.

I just checked and you are right, but this is definitely a bug. It is event float64, since the `subtract_pedestal_jit` function creates a new array without specifying the dtype. Will fix.

moralejo:
After the last commit, the waveform is converted to float32 in any case. So there is no longer any problem with the negative values, right? In that case we can remove the clipping of the correction, or am I overlooking something @SeiyaNozaki ?

maxnoe:
@moralejo No, I think when a pedestal file is created, the script directly calls the `time_lapse_corr` function, not the `apply_drs4_corrections` function. So it can still be uint16 there.

The fix I made only makes sure it's float32, not float64.

SeiyaNozaki:
Yes, this is also my understanding. Thank you, @maxnoe !
and please forget my last comment... I forgot the additional 400 offsets even after pedestal subtraction.
So the current fix should be fine!

moralejo:
> @moralejo No, I think when a pedestal file is created, the script directly calls the `time_lapse_corr` function, not the `apply_drs4_corrections` function. So it can still be uint16 there.
> 

Ok, but then, when processing data the time lapse correction also has the clipping, right?  I think it should not. Especially if the timelapse thing is applied after the pedestal baseline subtraction, because there will be a lot of (genuinely) negative values.

SeiyaNozaki:
If I understand correctly, all of the data should have positive values in this stage even after the pedestal subtraction because those data have additional 400 counts offset given by EVB. Then, these 400 counts are subtracted after all of the DRS corrections. So this clipping rarely works during normal data processing (R0 to DL1).

moralejo:
> If I understand correctly, all of the data should have positive values in this stage even after the pedestal subtraction because those data have additional 400 counts offset given by EVB. Then, these 400 counts are subtracted after all of the DRS corrections. So this clipping rarely works during normal data processing (R0 to DL1).

Ah, ok, I think you are right, "subtraction" stands for "equalization" (to a common baseline of 400) - then I agree, it won't harm (but would anyway be unnecessary in the case the waveform is float)

moralejo (approved):

PR #96, opened 2021-04-03
Spike subtraction instead of interpolation

SeiyaNozaki:
continuation of https://github.com/cta-observatory/cta-lstchain/pull/574 

This PR is to subtract corresponding ADC counts on spike cells instead of interpolation.
(related this issue: https://github.com/cta-observatory/cta-lstchain/issues/571)

In this PR, I added a function to find spike cells, put corresponding flags, and subtract spike ADC counts. 
Please find the attached summary of my checks using this new function.
[20210403_SpikeCorrection.pdf](https://github.com/cta-observatory/ctapipe_io_lst/files/6252797/20210403_SpikeCorrection.pdf)

For now, spike ADC counts are hard-coded, and common values are used in the camera.
But, it might be better to find spike heights for each pixel during the computation of drs pedestals.
(I'd like to hear your comments again @TomohikoOka @tysaito2008 @pawel21 @jsitarek )
I will make another PR in lstchain for that after https://github.com/cta-observatory/cta-lstchain/pull/601

### DIFF
--- a/ctapipe_io_lst/calibration.py
+++ b/ctapipe_io_lst/calibration.py
@@ -784,6 +784,42 @@ def find_spike_position_jit(
 
                     if (spike_pos - current_fc) % N_CAPACITORS_CHANNEL < N_SAMPLES:
                         spike_flag[gain, pixel, k, (spike_pos - current_fc) % N_CAPACITORS_CHANNEL] = True   
+
+@njit(cache=True)
+def find_spike_position_jit_data_from_20181010_to_20191104(
+    first_capacitor,
+    first_capacitor_old,
+    spike_flag,
+):
+    """
+    Numba function for finding spike positions in readout windows
+    """
+    LAST_IN_FIRST_HALF = N_CAPACITORS_CHANNEL // 2 - 1
+
+    for gain in range(N_GAINS):
+        for pixel in range(N_PIXELS):
+            last_fc = first_capacitor_old[gain, pixel] % N_CAPACITORS_CHANNEL
+            current_fc = first_capacitor[gain, pixel] % N_CAPACITORS_CHANNEL
+            last_lc = (last_fc + N_SAMPLES - 1) % N_CAPACITORS_CHANNEL
+            current_lc = (current_fc + N_SAMPLES - 1) % N_CAPACITORS_CHANNEL
+
+            spike_flag[gain, pixel] = np.zeros((3, N_SAMPLES), dtype='bool') 
+
+            # The correction is only needed for even
+            # last capacitor (lc) in the first half of the DRS4 ring
+            if last_lc % 2 == 0 and last_lc<= LAST_IN_FIRST_HALF:
+                
+                for k in range(3):
+                    spike_pos = (last_lc + k - 1) % N_CAPACITORS_CHANNEL
+
+                    if (spike_pos - current_fc) % N_CAPACITORS_CHANNEL < N_SAMPLES:
+                        spike_flag[gain, pixel, k, (spike_pos - current_fc) % N_CAPACITORS_CHANNEL] = True
+
+                    spike_pos = (N_CAPACITORS_CHANNEL - 3 + k - last_lc) % N_CAPACITORS_CHANNEL
+
+                    if (spike_pos - current_fc) % N_CAPACITORS_CHANNEL < N_SAMPLES:
+                        spike_flag[gain, pixel, k, (spike_pos - current_fc) % N_CAPACITORS_CHANNEL] = True   
+
                         
 @njit(cache=True)
 def ped_time(timediff):
--- a/ctapipe_io_lst/calibration.py
+++ b/ctapipe_io_lst/calibration.py
@@ -163,6 +163,7 @@ class LSTR0Corrections(TelescopeComponent):
         self.last_readout_time = {}
         self.first_cap = {}
         self.first_cap_old = {}
+        self.spike_flag = {}
         self.fbn = {}
         self.fan = {}
 
@@ -173,6 +174,9 @@ class LSTR0Corrections(TelescopeComponent):
             shape = (N_GAINS, N_PIXELS)
             self.first_cap[tel_id] = np.zeros(shape, dtype=int)
             self.first_cap_old[tel_id] = np.zeros(shape, dtype=int)
+
+            shape = (N_GAINS, N_PIXELS, 3, N_SAMPLES)
+            self.spike_flag[tel_id] = np.zeros(shape, dtype=bool)
 
         # set the right default for our default selector, change back afterwards
         # to not impact other code.
@@ -449,6 +453,7 @@ class LSTR0Corrections(TelescopeComponent):
 
         container.waveform = waveform
 
+
     def interpolate_spikes(self, event, tel_id):
         """
         Interpolates spike A & B.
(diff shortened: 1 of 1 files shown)


ctapipe_io_lst/calibration.py
```diff
@@ -200,7 +204,9 @@ def apply_drs4_corrections(self, event: ArrayEventContainer):
             # apply drs4 corrections
             self.subtract_pedestal(event, tel_id)
             self.time_lapse_corr(event, tel_id)
-            self.interpolate_spikes(event, tel_id)
```
jsitarek, line 203: maybe make here a switch, if interpolation or subtraction should be used

ctapipe_io_lst/calibration.py
```diff
@@ -574,6 +581,48 @@ def interpolate_pseudo_pulses_data_from_20181010_to_20191104(waveform, fc, fc_ol
...
+        # one for data from 2019/11/05 (from Run 1574) after update firmware.
+        # The old readout (before 2019/11/05) is shifted by 1 cell.
+        run_id = event.lst.tel[tel_id].svc.configuration_id        
+        if run_id > LAST_RUN_WITH_OLD_FIRMWARE:
+            subtract_spikes = subtract_spikes_jit
+        else:
+            subtract_spikes = subtract_spikes_jit_data_from_20181010_to_20191104
+
+        # corresponding spike adc counts are hard-coded for now. Those values will be loaded from a pedestal file.
+        spike_adc = np.array([47, 54, 8])
```
jsitarek, line 616: those hard coded values are indeed not very clear, why not implement reading them from the pedestal file immediatelly? 

also, do you have enough statistics in the general pedestal file to estimate those values, or is a special run needed?
if a special run is needed, and no night-to-night changes are expected, another possibility would be to pass those values via configuration file (with those values as defaults)

ctapipe_io_lst/calibration.py
```diff
@@ -719,6 +768,86 @@ def do_time_lapse_corr_data_from_20181010_to_20191104(
...
+            spike_flag[gain, pixel] = np.zeros((3, N_SAMPLES), dtype='bool') 
+
+            # The correction is only needed for even
+            # last capacitor (lc) in the first half of the DRS4 ring
+            if last_lc % 2 == 0 and last_lc<= LAST_IN_FIRST_HALF:
+                
+                for k in range(3):
+                    spike_pos = (last_lc + k) % N_CAPACITORS_CHANNEL
+
+                    if (spike_pos - current_fc) % N_CAPACITORS_CHANNEL < N_SAMPLES:
```
jsitarek, line 801: you use (spike_pos - current_fc) % N_CAPACITORS_CHANNEL in 3 lines, maybe call it a new variable

ctapipe_io_lst/calibration.py
```diff
@@ -719,6 +768,86 @@ def do_time_lapse_corr_data_from_20181010_to_20191104(
...
+            spike_flag[gain, pixel] = np.zeros((3, N_SAMPLES), dtype='bool') 
+
+            # The correction is only needed for even
+            # last capacitor (lc) in the first half of the DRS4 ring
+            if last_lc % 2 == 0 and last_lc<= LAST_IN_FIRST_HALF:
+                
+                for k in range(3):
+                    spike_pos = (last_lc -1 + k) % N_CAPACITORS_CHANNEL
+
+                    if (spike_pos - current_fc) % N_CAPACITORS_CHANNEL < N_SAMPLES:
```
jsitarek, line 841: same comment as above, you can introduce a new variable here since it is used 3 times

ctapipe_io_lst/calibration.py
```diff
@@ -163,6 +163,7 @@ def __init__(self, subarray, config=None, parent=None, **kwargs):
         self.last_readout_time = {}
         self.first_cap = {}
         self.first_cap_old = {}
+        self.spike_flag = {}
```
jsitarek, line 166: this is only used to tag the changed slices in the code, it is not used for actual calculation, right?

have you checked if the subtraction code and the old interpolation code are similar in speed?
removing those spike_flag tags could speed it up a bit more.

ctapipe_io_lst/calibration.py
```diff
@@ -719,6 +768,86 @@ def do_time_lapse_corr_data_from_20181010_to_20191104(
...
+            last_fc = first_capacitor_old[gain, pixel] % N_CAPACITORS_CHANNEL
+            current_fc = first_capacitor[gain, pixel] % N_CAPACITORS_CHANNEL
+            last_lc = (last_fc + N_SAMPLES - 1) % N_CAPACITORS_CHANNEL
+            current_lc = (current_fc + N_SAMPLES - 1) % N_CAPACITORS_CHANNEL
+
+            spike_flag[gain, pixel] = np.zeros((3, N_SAMPLES), dtype='bool') 
+
+            # The correction is only needed for even
+            # last capacitor (lc) in the first half of the DRS4 ring
+            if last_lc % 2 == 0 and last_lc<= LAST_IN_FIRST_HALF:
```
jsitarek, line 836: I thought that this condition on the last capacitor should be inverted for the data before and after Nov 2019, while it is the same one here, I checked the interpolation function and the condition there is also the same in both cases, @pawel21  and @SeiyaNozaki can you check that it is really intended like this (in both interpolation and subtraction)

jsitarek (changes requested):
I left a few comments, the only one that is possibly critical is the one about the last capacitor condition between the two subtraction functions. Please check that there is no mistake in the code here.
Other comments are just suggestions.

ctapipe_io_lst/calibration.py
maxnoe: See https://github.com/cta-observatory/cta-lstchain/pull/725

the standard 20k Events seems enough if we only need spike heights per pixel and not per capacitor.

ctapipe_io_lst/calibration.py
SeiyaNozaki: yep, I expected it like you at that time, but it seems the same condition is needed to predict spike position...
As of now, I don't understand why it is not inverted, but apply this condition because it seems actually correct...

SeiyaNozaki:
Hi @maxnoe,
Let me clarify one thing.
I guess you are developing the codes to subtract spike heights using the corresponding table generated with cta-observatory/cta-lstchain#725 (e.g. get_spike_A_positions method) with other code refactorings.
If so, should I stop here for this PR and review your codes later?

maxnoe:
@SeiyaNozaki Yes, I guess it makes sense to wait until at least #105 is merged, since that makes some relatively large refactoring needed for already gain selected data, which in turn also makes it easier to just implement a different spike correction algorithm.

We can discuss about how best to apply it and from which files and so on, but the splitting up of the large spike correction function into several independent ones makes it much easier. We need much fewer functions that way. 

Instead of all combinations of spike correction method, gain selection method, and firmware version we can just combine them.

ctapipe_io_lst/calibration.py
jsitarek: thx for checking @SeiyaNozaki 
either way this part of code is only for the old pre Nov-2019 data, and will likely be removed

jsitarek:
Hi @maxnoe 

about the conflict with #105  - what you split into several functions is the dt correction curve, the spike interpolation was already split. 
Nevertheless, since the changes in @105 are larger than in here, I guess it will be indeed simpler to first merge 105 and then update this one (or rather redo it from scratch, because likely updating this branch after the #105 merge will fail badly)

maxnoe:
@jsitarek I also split the spike correction into the loop over all gains / pixels part and the actual application to a single pixel. Next step is to split the position calculation from the correction application.

jsitarek:
ok @maxnoe, I was confused, because the old code also had the actual interpolation separated into a different function, but I see that you indeed added one more in the chain, i.e. LSTR0Corrections::interpolate_spikes ==> interpolate_spikes/interpolate_spikes_gain_selected ==> interpolate_spikes_pixel/ interpolate_spikes_pixel_old_firmware ==>  interpolate_spike_A Let me comment on this in the other PR

maxnoe:
Superseded by #121

PR #99, opened 2021-04-09
Skip events with event_id==0

maxnoe:
To address: https://github.com/cta-observatory/cta-lstchain/issues/684


morcuended:
The fix worked fine

morcuended (approved):

PR #102, opened 2021-04-29
Add options to disable delta t and spike correction

maxnoe:


jsitarek (commented):
Hi @maxnoe thx for this implementation.

the previous version of the code  applied dt correction and spike interpolation when a drs4 pedestal files was given (see the __init__.py file). 
Now with those switches it is possible that somebody explicitely asks for this correction (even while it is default) but do not specify the DRS4 file. 

@pawel21 should correct me if I'm wrong, but I think that in the current implementation for the time lapse correction and spike interpolation algorithms do not require any information in the DRS4 calibration file. And this actually also makes a lot of sense because basic subtraction of the baseline is already done in EVB and we may or may not want to redo it with a second file. 

Therefore I would suggest to either:
1) allow to run the DT correction and interpolation even without DRS4 file
or
2) add a consistency check that if those corrections are requested and no DRS4 calibration file is given raise an exception

maxnoe:
Good point, @jsitarek, right now, providing the DRS4 pedestal file is the general switch to say "I want corrections", independent of if these need that particular file.

Let me think about the best solution here is. The `appy_*` options right now are for the `LSTR0Corrections` class, not the event source itself, so these defaults are in the sense of "If we apply corrections in general, these corrections will be applied".

maxnoe:
I came up with the following:

  * Add option to LSTEventSource to completely enable/disable DRS4
  corrections
  
  * Add options for the different DRS4 calibration steps to
  LSTR0Corrections
  
  * Enable all corrections by default, raising error if needed calibration
  files are not given.


I think that's what you had in mind, right @jsitarek ?

Please have a look, also pinging @FrancaCassol @moralejo

ctapipe_io_lst/calibration.py
```diff
@@ -240,7 +255,12 @@ def calibrate(self, event: ArrayEventContainer):
 
         for tel_id in event.r0.tel:
             r1 = event.r1.tel[tel_id]
-            waveform = r1.waveform
+            # if `apply_drs4_corrections` is False, we did not fill in the
+            # waveform yet.
+            if r1.waveform is None:
+                waveform = event.r0.tel[tel_id].waveform.astype(np.float32)
```
jsitarek, line 261: I'm wondering about this line, since it was not there in the previous code. Did you found an actual example when this waveform was missing in r1? Nevertheless, the line looks perfectly safe for me.

maxnoe: Before, there was no option to not do any DRS4 corrections before applying this calibration and the R1 would always be filled by the `apply_drs4_corrections` function. No we have the option to completely disable drs4 corrections, and in that case, r1 will not yet be filled at this point.

jsitarek (approved):
I went through the changes and they seem fine to me

FrancaCassol (approved):

PR #105, opened 2021-06-22
Implement support for already gain selected input data

maxnoe:
Reading in gain selected R1 written the tool here: 

https://gitlab.cta-observatory.org/cta-array-elements/lst/subsystems/daq/offline_gain_selection

This now works.

Comparisons to before coming.

### DIFF
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -644,7 +644,7 @@ class LSTEventSource(EventSource):
             reordered_waveform = np.full((N_PIXELS, N_SAMPLES), fill, dtype=dtype)
             reordered_waveform[expected_pixels] = waveform
 
-            reordered_selected_gain = np.zeros(N_PIXELS, dtype=np.uint8)
+            reordered_selected_gain = np.full(N_PIXELS, -1, dtype=np.int8)
             reordered_selected_gain[expected_pixels] = selected_gain
 
             r0 = R0CameraContainer()
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -611,7 +611,7 @@ class LSTEventSource(EventSource):
     def fill_r0r1_camera_container(self, zfits_event):
         """
         Fill the r0 or r1 container, depending on whether gain
-        selection has already happened (r0) or not (r1)
+        selection has already happened (r1) or not (r0)
 
         This will create waveforms of shape (N_GAINS, N_PIXELS, N_SAMPLES),
         or (N_PIXELS, N_SAMPLES) respectively regardless of the n_pixels, n_samples
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -61,6 +61,12 @@ class TriggerBits(IntFlag):
 
 
 class PixelStatus(IntFlag):
+    '''
+    Pixel status information
+
+    See Section A.5 of the CTA R1 Data Model:
+    https://forge.in2p3.fr/dmsf/files/8627
+    '''
     RESERVED_0 = auto()
     RESERVED_1 = auto()
     HIGH_GAIN_STORED = auto()
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -139,7 +139,9 @@ def read_pulse_shapes():
 
 
 class LSTEventSource(EventSource):
-    """EventSource for LST r0 data."""
+    """
+    EventSource for LST R0 data.
+    """
     multi_streams = Bool(
         True,
         help='Read in parallel all streams '
@@ -201,6 +203,18 @@ class LSTEventSource(EventSource):
     classes = [PointingSource, EventTimeCalculator, LSTR0Corrections]
 
     def __init__(self, input_url=None, **kwargs):
+        '''
+        Create a new LSTEventSource.
+
+        Parameters
+        ----------
+        input_url: Path
+            Path to or url understood by ``ctapipe.core.traits.Path``.
+            If ``multi_streams`` is ``True``, the source will try to read all
+            streams matching the given ``input_url``
+        **kwargs:
+            Any of the traitlets. See ``LSTEventSource.class_print_help``
+        '''
         super().__init__(input_url=input_url, **kwargs)
 
         if self.multi_streams:
--- a/ctapipe_io_lst/__init__.py
+++ b/ctapipe_io_lst/__init__.py
@@ -20,8 +20,10 @@ from enum import IntFlag, auto
 
 from ctapipe.io import EventSource
 from ctapipe.io.datalevels import DataLevel
-from ctapipe.core.traits import Int, Bool, Float, Enum
-from ctapipe.containers import PixelStatusContainer, EventType
+from ctapipe.core.traits import Bool, Float, Enum
+from ctapipe.containers import (
+    PixelStatusContainer, EventType, R0CameraContainer, R1CameraContainer,
+)
 
 from .containers import LSTArrayEventContainer, LSTServiceContainer
 from .version import __version__
@@ -36,7 +38,7 @@ from .anyarray_dtypes import (
     TIB_DTYPE,
 )
 from .constants import (
-    HIGH_GAIN, N_PIXELS, N_SAMPLES
+    HIGH_GAIN, N_GAINS, N_PIXELS, N_SAMPLES
 )
 
 __all__ = ['LSTEventSource', '__version__']
@@ -57,6 +59,18 @@ class TriggerBits(IntFlag):
     PHYSICS = MONO | STEREO
     OTHER = CALIBRATION | SINGLE_PE | SOFTWARE | PEDESTAL | SLOW_CONTROL
 
+
+class PixelStatus(IntFlag):
+    RESERVED_0 = auto()
+    RESERVED_1 = auto()
+    HIGH_GAIN_STORED = auto()
+    LOW_GAIN_STORED = auto()
+    SATURATED = auto()
+    PIXEL_TRIGGER_1 = auto()
+    PIXEL_TRIGGER_2 = auto()
+    PIXEL_TRIGGER_3 = auto()
+
+    BOTH_GAINS_STORED = HIGH_GAIN_STORED | LOW_GAIN_STORED
 
 OPTICS = OpticsDescription(
     'LST',
@@ -126,12 +140,6 @@ def read_pulse_shapes():
 
 class LSTEventSource(EventSource):
     """EventSource for LST r0 data."""
-
-    n_gains = Int(
-        2,
-        help='Number of gains at r0 level'
-    ).tag(config=True)
-
     multi_streams = Bool(
         True,
         help='Read in parallel all streams '
@@ -193,22 +201,6 @@ class LSTEventSource(EventSource):
     classes = [PointingSource, EventTimeCalculator, LSTR0Corrections]
 
     def __init__(self, input_url=None, **kwargs):
-        """
-        Parameters
-        ----------
-        n_gains = number of gains expected in input file
-
-        multi_streams = enable the reading of input files from all streams
-
-        config: traitlets.loader.Config
-            Configuration specified by config file or cmdline arguments.
-            Used to set traitlet values.
-            Set to None if no configuration to pass.\
-        kwargs: dict
-            Additional parameters to be passed.
-            NOTE: The file mask of the data to read can be passed with
-            the 'input_url' parameter.
-        """
         super().__init__(input_url=input_url, **kwargs)
 
         if self.multi_streams:
@@ -218,7 +210,7 @@ class LSTEventSource(EventSource):
 
             path, name = os.path.split(os.path.abspath(self.input_url))
             if 'Run' in name:
-                stream, run = name.split('Run', 1)
+                _, run = name.split('Run', 1)
             else:
                 run = name
 
@@ -344,16 +336,9 @@ class LSTEventSource(EventSource):
                 self.log.warning('Event with event_id=0 found, skipping')
                 continue
 
-            # fill general R0 data
-            self.fill_r0_container(array_event, zfits_event)
-
-            # fill specific LST event data
+            self.fill_r0r1_container(array_event, zfits_event)
             self.fill_lst_event_container(array_event, zfits_event)
-
-            # fill trigger info (needs r0 and lst specifics)
             self.fill_trigger_info(array_event)
-
-            # fill general monitoring data
             self.fill_mon_container(array_event, zfits_event)
             self.fill_pointing_info(array_event)
 
@@ -367,11 +352,10 @@ class LSTEventSource(EventSource):
 
             # gain select and calibrate to pe
             if self.r0_r1_calibrator.calibration_path is not None:
-
                 # skip flatfield and pedestal events if asked
                 if (
-                    array_event.trigger.event_type not in {EventType.FLATFIELD, EventType.SKY_PEDESTAL}
-                        or self.calibrate_flatfields_and_pedestals
+                    self.calibrate_flatfields_and_pedestals
+                    or array_event.trigger.event_type not in {EventType.FLATFIELD, EventType.SKY_PEDESTAL}
                 ):
                     self.r0_r1_calibrator.calibrate(array_event)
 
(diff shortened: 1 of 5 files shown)


maxnoe:
There is one thing left I need to fix: with the current implementation in this branch, the last readout times are no longer properly updated in the gain selected case. We also need to update the last readout times of the not selected gain.

maxnoe:
Ok, this is fixed now. 

I compared the results for subrun 2610.0000 for all four streams three way:

a) Current master, input not gain selected
b) This branch, input not gain selected
c) This branch, input gain selected using https://gitlab.cta-observatory.org/cta-array-elements/lst/subsystems/daq/offline_gain_selection

a) and b) are always exactly identical, so there is no behaviour change between this branch and master for not gain selected data, at least not on the file I tested. (small test files and the 2610.0000 run).

c) differs only in two waveforms out of the 53k times 1855 pixels, in which the C++ tool chose the low gain channel and `ctapipe_io_lst` chose the high gain. I suspect this is due to the drs4 calibration, since that is not applied in the C++ tool before doing the gain selection.

I think we can live with this.


On another note: do we need to keep the code for reading data with the old firmware? It would get rid of quite a lot of duplicated code (cdts dtypes, updating last readout) if we would not try to keep compatibility with the data before 2019-11.

Any opinions? @moralejo @rlopezcoto @FrancaCassol @jsitarek @pawel21

maxnoe:
Some design decisions:

* If the input is already gain selected, R0 is not filled, but directly R1 from the gain selected waveform and the `selected_gain_channel` is filled from the `pixel_status` flags
* The code currently assumes that either all or no pixels where gain selected. There is no support for partial gain selection.

moralejo:
> 
> On another note: do we need to keep the code for reading data with the old firmware? It would get rid of quite a lot of duplicated code (cdts dtypes, updating last readout) if we would not try to keep compatibility with the data before 2019-11.
> 
> Any opinions? @moralejo @rlopezcoto @FrancaCassol @jsitarek @pawel21

The oldest sky data we process are from Nov 23, 2019. **From my side** I'd say we can clean the code as you propose. In case reading older data is needed, one can use the current latest release of io_lst. 
If anyone prefers to keep the possibility of reading those old data, please speak up as soon as possible.

jsitarek:
Hi @maxnoe,

thx for the implementation. One question, when you compared the cases a), b), c)  you checked only the waveforms, right? I think it would be good to check also the reconstructed times, because it is easy to mess up the time calibration during the gain selection, if e.g. the first capacitor values  are taken from wrong gain. Maybe you could also implement a test with comparing the time calibration from non-gain-selected with gain-selected files?

maxnoe:
>  I think it would be good to check also the reconstructed times, because it is easy to mess up the time calibration during the gain selection, if e.g. the first capacitor values are taken from wrong gain. Maybe you could also implement a test with comparing the time calibration from non-gain-selected with gain-selected files?

When you say "reconstructed times", do you mean the `last_readout_time` or the DL1 peak time? I assume the former, since the latter cannot really change if we assert that the waveforms are exactly identical.

That's a good test, I will add it to the unit test comparing the two results.

jsitarek:
Hi,

I've actually meant reconstructed time. It depends mainly on the waveforms as you say, but the correction is also using the first capacitor, that is different in each pixel for low and high gain channel, hence I suggested the test. 

The last reading time you can check as well, but I think you are already checking the time-lapse pedestal correction, and this one depends on the last reading time, so I think it should be fine,

maxnoe:
@jsitarek Ha, right, the calculated correction for the DL1 event, yes. 

There is no change to that in this PR, as the code needed to work with gain selected data already before (when the gain selection is performed in LSTR0Corrections before calculating the DL1 time correction). But I can add the test.

maxnoe:
@jsitarek @FrancaCassol @moralejo 

I tested this branch now also with the gain selected file written by EVB: `/fefs/aswg/data/real/R0/20210319/LST-1.1.Run04164.0000.fits.fz` and this branch is able to read the file.

moralejo:
Great, thanks a lot @maxnoe.
Fine from my side, but I'd prefer that @jsitarek or @FrancaCassol give the formal PR approval.

ctapipe_io_lst/__init__.py
```diff
@@ -617,44 +608,69 @@ def fill_pointing_info(self, array_event):
                 'No drive report specified, pointing info will not be filled'
             )
 
-    def fill_r0_camera_container(self, r0_container, zfits_event):
+    def fill_r0r1_camera_container(self, zfits_event):
         """
-        Fill with R0CameraContainer
+        Fill the r0 or r1 container, depending on whether gain
+        selection has already happened (r0) or not (r1)
```
jsitarek, line 614: isn't it the other way around?
the gain_selected data are written in r1

ctapipe_io_lst/__init__.py
```diff
@@ -726,20 +740,17 @@ def __init__(self, file_list):
...
-                    if self.camera_config is None:
-                        self.camera_config = self._camera_config[path]
-
             except StopIteration:
                 pass
 
-        # verify that somewhere the CameraConfing is present
-        if self.camera_config is None:
+        # verify that we found a CameraConfig
+        if len(self._camera_config) == 0:
```
jsitarek, line 750: it allows for a situation that some of the files are missing CameraConfig, or that the CameraConfig between different files are not matching, is it intended?
maybe check that the len of camera configs is the same as len of the file list

ctapipe_io_lst/calibration.py
```diff
@@ -211,15 +211,20 @@ def __init__(self, subarray, config=None, parent=None, **kwargs):
...
+        for tel_id, r0 in event.r0.tel.items():
             r1 = event.r1.tel[tel_id]
             # fill r1 waveform with a copy of the r0 converted to float32
             # float32 can represent all values of uint16 exactly, so this
             # does not loose precision.
-            r1.waveform = event.r0.tel[tel_id].waveform.astype(np.float32)
+            if r1.waveform is None:
+                r1.waveform = r0.waveform
+            else:
+                r1.waveform = r1.waveform
```
jsitarek, line 225: why do you make a=a type asignment?

ctapipe_io_lst/calibration.py
```diff
@@ -500,123 +520,191 @@ def interpolate_spikes(self, event, tel_id):
...
+def convert_to_pe(waveform, calibration, selected_gain_channel):
+    if selected_gain_channel is None:
+        waveform -= calibration.pedestal_per_sample[:, :, np.newaxis]
+        waveform *= calibration.dc_to_pe[:, :, np.newaxis]
+    else:
+        waveform -= calibration.pedestal_per_sample[selected_gain_channel, PIXEL_INDEX, np.newaxis]
+        waveform *= calibration.dc_to_pe[selected_gain_channel, PIXEL_INDEX, np.newaxis]
+
+@njit(cache=True)
+def interpolate_spike_A(waveform, position):
```
jsitarek, line 550: I think this function was not changed, why did you move it so it shows up in the diff?

ctapipe_io_lst/calibration.py
```diff
@@ -500,123 +520,191 @@ def interpolate_spikes(self, event, tel_id):
...
+    Change waveform array.
+    """
+    a = int(waveform[position - 1])
+    b = int(waveform[position + 2])
+    waveform[position] = waveform[position - 1] + (0.33 * (b - a))
+    waveform[position + 1] = waveform[position - 1] + (0.66 * (b - a))
+
+
+@njit(cache=True)
+def interpolate_spikes_pixel(waveform, current_fc, last_fc):
```
jsitarek, line 562: same thing here, you changed the name of the function and parameters, but body of the function does not seem to be much different, it makes the diff quite messy

jsitarek (changes requested):
thx @maxnoe for the development

the only problematic thing that I saw was that the same thresholds are used for selecting ff events no matter if the pixel is in HG or LG

A few smaller comments also included.

not all of the function renamings were needed, and they will complicate to merge the PR #96
also some of the functions were moved and thus show up in diff, I didn't check them too carefully, so hopefully there were no mistakes there.

ctapipe_io_lst/__init__.py
```diff
@@ -193,22 +201,6 @@ class LSTEventSource(EventSource):
     classes = [PointingSource, EventTimeCalculator, LSTR0Corrections]
 
     def __init__(self, input_url=None, **kwargs):
-        """
-        Parameters
```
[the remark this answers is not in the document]
maxnoe: This was duplicated (incomplete and outdated) with the `help` strings of the traitlets.

ctapipe_io_lst/__init__.py
```diff
@@ -581,11 +565,18 @@ def tag_flatfield_events(self, array_event):
...
+        waveform = array_event.r1.tel[tel_id].waveform
+
+        # needs to work for gain already selected or not
+        if waveform.ndim == 3:
+            image = waveform[HIGH_GAIN].sum(axis=1)
+        else:
+            image = waveform.sum(axis=1)
 
-        image = array_event.r1.tel[tel_id].waveform[HIGH_GAIN].sum(axis=1)
         in_range = (image >= self.min_flatfield_adc) & (image <= self.max_flatfield_adc)
```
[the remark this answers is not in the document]
maxnoe: Flatfield events will not be gain selected (neither in EVB nor in the cpp tool I wrote).

ctapipe_io_lst/__init__.py
```diff
@@ -581,11 +565,18 @@ def tag_flatfield_events(self, array_event):
...
+        n_in_range = np.count_nonzero(in_range)
 
-        if np.count_nonzero(in_range) >= self.min_flatfield_pixel_fraction * image.size:
+        if n_in_range >= self.min_flatfield_pixel_fraction * image.size:
```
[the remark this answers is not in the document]
maxnoe: readability

ctapipe_io_lst/__init__.py
maxnoe: yes, good catch.

ctapipe_io_lst/__init__.py
maxnoe: The logic of the check has not changed. I just improved the way it was done. And yes, I think the intention was to allow missing camera configs in streams and only checking that at least one stream has a camera config.

ctapipe_io_lst/calibration.py
maxnoe: unnecessary, correct.

ctapipe_io_lst/__init__.py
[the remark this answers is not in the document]
jsitarek: agree for multi_streams and config, but kwargs and input_url you could have kep

ctapipe_io_lst/__init__.py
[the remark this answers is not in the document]
jsitarek: I know that this is the ultimate plan, but I'm wondering if this is working already like this. Do you know if the trigger type is known in EVB (and only later on when the files are written by EVB this information gets somehow corrupted), if not the calibration events cannot be written reliably with both gains if rest of the events is written with gain selection.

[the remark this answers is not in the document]
maxnoe: We talked to the EVB guys, gain selection for regular data written by EVB will not be done until this is reliable (and it also requires a data model change to encode the preprocessing options).

So for now: the cpp tool for the offline gain selection performs the same checks as are done here, so we are safe.

For the future: once EVB really starts writing gain selected or DRS4 preprocessed data, we anyway need to make changes again (including updating the protozfits python bindings)

ctapipe_io_lst/__init__.py
maxnoe: I'd say this is how it was intended and it's not really connected to this PR.

ctapipe_io_lst/calibration.py
maxnoe: It was `waveform = r1.waveform` before, when refactoring, I didn't realize that the statement was no superflous

ctapipe_io_lst/calibration.py
maxnoe: It was down with functions related to the timelapse correction, I preferred all the spike related functions closely together, in the order they are needed to be defined (this is used by the next function, which is used by the next function...)

ctapipe_io_lst/calibration.py
maxnoe: I broke this function out of the large function also containing the loops, because this is the same whether gain selection has already happend or not, only the loop is different. No real way to avoid the not-so-nice diff I think

maxnoe:
@jsitarek I have fixed the two smaller issues. I am afraid I can't do much do make the diff nicer, since I really had to break out / change these functions.

jsitarek (approved):
Hi @maxnoe 
I resolved the discussions on the points where we converged. The only one left is this docstring in init, (i.e. to keep there the kwargs and input_url note), but it is a minor issue, so if you prefer to remove it either way, it is also fine with me, I'm marking it as "approved"

maxnoe:
@jsitarek I am already working on adding it back

jsitarek:
Hi @maxnoe 

I noticed that the spike interpolation is now a cascade of 4 functions:  LSTR0Corrections::interpolate_spikes ==> interpolate_spikes/interpolate_spikes_gain_selected ==> interpolate_spikes_pixel/ interpolate_spikes_pixel_old_firmware ==>  interpolate_spike_A, which is a bit messy and can lead also to some overhead. 

interpolate_spikes_pixel_old_firmware we might eventually throw away, but it still means that there is a cascade of 4 functions for a relatively simple task, 

One possible solution is to get rid of interpolate_spikes/interpolate_spikes_gain_selected and put this code in  LSTR0Corrections::interpolate_spikes, actually instead of copying both functions there you could even make the function much shorter by something like this:

for pixel in range(N_PIXELS):
    if r1.selected_gain_channel is None:
        gains=range(N_GAINS)
    else:
        gains=[r1.selected_gain_channel]
    for gain in gains:
            current_fc = first_capacitors[gain, pixel]
            last_fc = previous_first_capacitors[gain, pixel]

            if run_id > LAST_RUN_WITH_OLD_FIRMWARE:
                interpolate_spikes_pixel(
                    waveform=waveform[gain, pixel],
                    current_fc=current_fc,
                    last_fc=last_fc,
                )
            else:
                interpolate_spikes_pixel_old_firmware(
                    waveform=waveform[gain, pixel],
                    current_fc=current_fc,
                    last_fc=last_fc,
                )

this would simplify the code considerably (but on the other hand it might end up with worse overhead with gain selection)
what do you think @maxnoe ?

maxnoe:
> which is a bit messy and 

I don't think having small functions that do a single task is messy. Quite the opposite. 

> can lead also to some overhead.

No, this is LLVM compiled code via numba. I would bet some money (and we can check using numba's ability to print the machine code) that LLVM will inline those little functions down the line.

>function much shorter by something like this:
This approach would probably (I did not test it) harm performance much more than the functions, since now you iterate over the array in a very cache-unfriendly way for the non-gain-selected case.

jsitarek:
> > which is a bit messy and
> 
> I don't think having small functions that do a single task is messy. Quite the opposite.
> 
one small function per small task if perfectly fine, but here you have 4 in total :-). 

> > can lead also to some overhead.
> 
> No, this is LLVM compiled code via numba. I would bet some money (and we can check using numba's ability to print the machine code) that LLVM will inline those little functions.
> 
> > function much shorter by something like this:
> 
> This approach would probably (I did not test it) harm performance much more than the functions, since now you iterate over the array in a very cache-unfriendly way for the non-gain-selected case.

about the performance as I wrote above I was not sure, my main point was readibility of the code. Having to dig thorugh 4 functions (two of which are in two variants) to the actual interpolation is excessive in my opinion, that's why I suggested a solution that would skip one step and two functions in total (you could probably think of one that is more efficient CPU-wise). The fact that the same name is used for the function in LSTR0Corrections::interpolate_spikes and global interpolate_spikes can also lead to some confusion.

maxnoe:
> one small function per small task if perfectly fine, but here you have 4 in total :-).

Four small ones instead of one giant ;)

> The fact that the same name is used for the function in LSTR0Corrections::interpolate_spikes and global interpolate_spikes can also lead to some confusion.

Naming is hard. I'll see that I make the names more clear, but I really think having the logic split up in multiple functions is better. This will also make it easier to implement both the subtraction and the interpolation to be able to compare them.

I will have a look at the code again and see if I can improve. But just putting it in one big function with complex control flow is not going to be the solution.

jsitarek:
I completely agree that I cannot be one big function either, I was just wondering if one of the steps can be skipped. Just have a look if you can find a better solution, and if not it can stay like this (I already put approval on the PR)

ctapipe_io_lst/__init__.py
```diff
@@ -58,6 +60,18 @@ class TriggerBits(IntFlag):
...
+class PixelStatus(IntFlag):
+    RESERVED_0 = auto()
+    RESERVED_1 = auto()
+    HIGH_GAIN_STORED = auto()
+    LOW_GAIN_STORED = auto()
+    SATURATED = auto()
+    PIXEL_TRIGGER_1 = auto()
+    PIXEL_TRIGGER_2 = auto()
+    PIXEL_TRIGGER_3 = auto()
+
```
FrancaCassol, line 72: Hi @maxnoe,
to what do the pixel_trigger_1,2,3 correspond? (perhaps a comment?)

maxnoe: This is explained in the R1 document, https://forge.in2p3.fr/dmsf/files/8627, section A.5:

> bits 5-7" the last 3 bits are for recording pixel-wise (or sector-wise) trigger info (contents to be
determined)

I will add a link to the document.

maxnoe: For current LST data, they seem to always be 0

FrancaCassol (approved):

moralejo (approved):

moralejo:
We can move on with merging this, right? 
The data reducer and its reading back from file are already tested - or is there any other test that we need before starting to reduce data, @maxnoe?

PR #106, opened 2021-07-09
Fix dtype of dragon time, keep info about jumps

maxnoe:
Because of numpy's type promotion rules and the dtypes involved, `dragon_time` was actually a float64, not uint64, loosing us precision, I don't know how important that loss of precision was though.

@rlopezcoto How important might this be for the pulsar analysis?

### DIFF
--- a/ctapipe_io_lst/event_time.py
+++ b/ctapipe_io_lst/event_time.py
@@ -216,8 +216,8 @@ class EventTimeCalculator(TelescopeComponent):
                 run_summary = read_run_summary(self.run_summary_path.tel[tel_id])
                 row = run_summary.loc[run_id]
                 self._has_dragon_reference[tel_id] = True
-                self._dragon_reference_time[tel_id] = row['dragon_reference_time']
-                self._dragon_reference_counter[tel_id] = row['dragon_reference_counter']
+                self._dragon_reference_time[tel_id] = np.uint64(row['dragon_reference_time'])
+                self._dragon_reference_counter[tel_id] = np.uint64(row['dragon_reference_counter'])
                 self._dragon_module_index[tel_id] = row['dragon_reference_module_index']
 
                 if row['dragon_reference_source'] == 'run_start':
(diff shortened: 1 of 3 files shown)


rlopezcoto:
Hi @maxnoe, thanks for spotting this. In principle for Crab pulsar analysis we need a precision of the order of 100 us. I think this is also the order of magnitude needed for MAGIC+LST coincidences, so it should not have caused too much trouble.

maxnoe:
@rlopezcoto I checked and the maximum difference is 120 nanoseconds between the fixed version here and the one in master.

However with this here, we get the real ns precision (astropy iso format with ns precision agrees exactly with the uint64 values), this was not the case before.

The differences are distributed uniformly between -120ns and 120ns, if I'd scratch my head harder, I could probably explain that with the floating point precision.

ctapipe_io_lst/event_time.py
```diff
@@ -33,26 +31,39 @@ def validate(self, obj, value):
...
 
 
 def combine_counters(pps_counter, tenMHz_counter):
     '''
     Combines the values of pps counter and tenMHz_counter
     and returns the sum in ns.
     '''
-    return int(1e9) * pps_counter + 100 * tenMHz_counter
+    return (
+        S_TO_NS * pps_counter.astype(np.uint64)
```
rlopezcoto, line 64: instead of using variables like `S_TO_NS` or `TEN_MHZ_TO_NS` could it be a good idea to use astropy units or may that complicate things innecessarily?

ctapipe_io_lst/tests/test_event_time.py
```diff
@@ -11,13 +11,58 @@
...
-
-
 test_data = Path(os.getenv('LSTCHAIN_TEST_DATA', 'test_data'))
 test_run_summary = test_data / 'real/monitoring/RunSummary/RunSummary_20200218.ecsv'
 
 
+def test_combine_counters():
+    from ctapipe_io_lst.event_time import combine_counters
+
+    pps_counter = np.uint16(60_000)
```
rlopezcoto, line 21: just curiosity... is this name of writing integers coming from some pep recommendation?

rlopezcoto (commented):
> @rlopezcoto I checked and the maximum difference is 120 nanoseconds between the fixed version here and the one in master.
However with this here, we get the real ns precision (astropy iso format with ns precision agrees exactly with the uint64 values), this was not the case before.
The differences are distributed uniformly between -120ns and 120ns, if I'd scratch my head harder, I could probably explain that value with the floating point precision.

thanks @maxnoe! so the differences are not going to result in any change in any of the analysis performed and since now the astropy iso format is exact, I would say that this is fine.

ctapipe_io_lst/event_time.py
maxnoe: astropy units by default convert everything to float64. We really need uint64 values here. 

The astropy units code would be much more verbose to ensure we have the right dtypes and I don't think it would really benefit us here.

ctapipe_io_lst/tests/test_event_time.py
maxnoe: I am not sure I understand. Do you mean the underscore? That's just for readability, it was introduced in 3.6.

rlopezcoto: yes, I meant the underscore, thanks!

ctapipe_io_lst/event_time.py
rlopezcoto: good, I also think it should be ok to carry on as it is now

PR #108, opened 2021-09-28
Spike positions

maxnoe:
This adds functions to obtain the positions of the spikes needed for calculating the spike heights to enable spike subtraction instead of interpolation and a function to update the last_readout_times needed to create the time calibration file in this lstchain PR here: https://github.com/cta-observatory/cta-lstchain/pull/725

### DIFF
--- a/ctapipe_io_lst/calibration.py
+++ b/ctapipe_io_lst/calibration.py
@@ -556,8 +556,10 @@ def interpolate_spike_A(waveform, position):
 
 
 @njit(cache=True)
-def interpolate_spikes_pixel(waveform, current_fc, last_fc):
+def get_spike_A_positions(current_fc, last_fc):
     LAST_IN_FIRST_HALF = N_CAPACITORS_CHANNEL // 2 - 1
+
+    positions = []
 
     for k in range(4):
         # looking for spike A first case
@@ -570,7 +572,7 @@ def interpolate_spikes_pixel(waveform, current_fc, last_fc):
             # DRS4 ring
             last_capacitor = (last_fc + N_SAMPLES - 1) % N_CAPACITORS_CHANNEL
             if last_capacitor % 2 == 0 and last_capacitor <= LAST_IN_FIRST_HALF:
-                interpolate_spike_A(waveform, spike_A_position)
+                positions.append(spike_A_position)
 
         # looking for spike A second case
         abspos = N_SAMPLES - 1 + last_fc + k * N_CAPACITORS_CHANNEL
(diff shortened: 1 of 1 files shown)


ctapipe_io_lst/calibration.py
```diff
@@ -556,73 +556,108 @@ def interpolate_spike_A(waveform, position):
...
 
+    positions = []
+
     for k in range(4):
         # looking for spike A first case
         abspos = N_CAPACITORS_CHANNEL + 1 - N_SAMPLES - 2 - last_fc + k * N_CAPACITORS_CHANNEL + N_CAPACITORS_PIXEL
         spike_A_position = (abspos - current_fc + N_CAPACITORS_PIXEL) % N_CAPACITORS_PIXEL
 
-        if 2 < spike_A_position < (N_SAMPLES - 2):
+        # a spike affects the poistion itself and the two following slices
```
FrancaCassol, line 584: typo: "position"

FrancaCassol (approved):

ctapipe_io_lst/calibration.py
maxnoe: fixed

maxnoe:
@jsitarek sorry, I don't see any comment, can you post it here again or open an issue?

jsitarek:
I started the review and see it above, maybe you do not see it because I did not finalize the review. Either way this is my comment (to line 586):
"

this looks suspicious to me (there are 2 more similar changes below). First of all spike_A_position comes from modulo division, so it cannot be -1, hence the first condition will always be true.
Also, while for spike subtraction it should be fine, if the interpolation is done it has to be modified as well. You cannot interpolate the first slice (you can only extrapolate it by taking the value of the 2nd slice), also if the interpolation is done before subtraction of the noisy slices this will cause further problems. If I remember correcty first 3 slices and the last one slice is noisy, so they should not be used for interpolation.

@pawel21 can you also have a look at this?
"

maxnoe:
> I started the review and see it above, maybe you do not see it because I did not finalize the review. Either way this is my comment (to line 586):

Yes you need to submit it. Others cannot see "pending" comments.

maxnoe:
> You cannot interpolate the first slice (you can only extrapolate it by taking the value of the 2nd slice)

The calculation of the positions is independent of the application. The interpolation code checks that the calculated position is > 2, as before.

maxnoe:
But you are right about the negative issue with modulus... I assumed C like behaviour (where -2 % 1024 == -2)

PR #109, opened 2021-10-29
Multifile fixes

maxnoe:
Add context managers and move MultiFile to own module

* Add `__enter__`, `__exit__` and close methods to MultiFile
* Call `MultiFile.close` in `LSTEventSource.close`

### DIFF
--- a/ctapipe_io_lst/tests/test_multifile.py
+++ b/ctapipe_io_lst/tests/test_multifile.py
@@ -12,12 +12,35 @@ def test_multifile():
     paths = test_r0_dir.glob('LST-1.*.Run02005.0000_first50.fits.fz')
 
     with MultiFiles(paths) as multi_files:
+
+        assert len(multi_files) == 200
+        assert multi_files.num_inputs() == 4
+
         event_count = 0
         for event in multi_files:
             event_count += 1
             assert event.event_id == event_count
 
         assert event_count == 200
+
+
+def test_rewind():
+    from ctapipe_io_lst.multifiles import MultiFiles
+
+    paths = test_r0_dir.glob('LST-1.*.Run02005.0000_first50.fits.fz')
+
+    with MultiFiles(paths) as multi_files:
+
+        for expected_event_id in range(1, 10):
+            event = next(multi_files)
+            assert event.event_id == expected_event_id
+
+        # and again
+        multi_files.rewind()
+        for expected_event_id in range(1, 10):
+            event = next(multi_files)
+            assert event.event_id == expected_event_id
+
 
 def test_different_runs():
     from ctapipe_io_lst.multifiles import MultiFiles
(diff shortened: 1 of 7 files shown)


ctapipe_io_lst/tests/test_multifile.py
```diff
@@ -0,0 +1,54 @@
...
+
+        assert len(multi_files) == 200
+        assert multi_files.num_inputs() == 4
+
+        event_count = 0
+        for event in multi_files:
+            event_count += 1
+            assert event.event_id == event_count
+
+        assert event_count == 200
```
jsitarek, line 24: I think this last check is redundant, since you already check that len(multi_files)=200, and you have a simple for loop above increasing a local variable this 200 times, I think no matter what MultiFiles does, you will always have event_count set to 200 at the end of the loop

ctapipe_io_lst/tests/test_multifile.py
```diff
@@ -0,0 +1,54 @@
...
+def test_rewind():
+    from ctapipe_io_lst.multifiles import MultiFiles
+
+    paths = test_r0_dir.glob('LST-1.*.Run02005.0000_first50.fits.fz')
+
+    with MultiFiles(paths) as multi_files:
+
+        for expected_event_id in range(1, 10):
+            event = next(multi_files)
+            assert event.event_id == expected_event_id
```
jsitarek, line 36: this line in principle could be skipped as well, since it is not testing the rewinding, but rather normal reading of the file (and exactly the same test is done in the function above)

jsitarek (approved):
I do not see any problems. 
In the tests, two lines I think can be removed because they seem redundant to me, but if you want to keep them it is also perfectly fine with me.

ctapipe_io_lst/tests/test_multifile.py
maxnoe: No, it's not redundant. Since `MultiFile` implements the `iterator` interface, `__len__` can do something different than iterating over it. Checking both tests the implementation of `__len__` and of `__iter__` and `__next__`

ctapipe_io_lst/tests/test_multifile.py
maxnoe: It checks that after "rewinding" the normal iteration works as expected. This is clearly not redundant, this is is exactly what failed before I fixed "rewind"

jsitarek: please check my comment again, I've explicitly meant this line 36 and not the whole function. You read first 10 events checking that the numbers increase, then rewind and then read them again checking again the numbers. The second check (in line 42) is perfectly valid, but if assert of line 36 fails it would also fail in line 22, hence I think you could just read the 10 events (without checking) then rewind and read them again (this time with checking). But as I wrote above this is just a redundant line, so if you want to keep it, it is fine with me

ctapipe_io_lst/tests/test_multifile.py
jsitarek: ok, thx for explanation, resolving comment

PR #117, opened 2021-11-08
Add separate option for the offset of the drs4 pedestal values

maxnoe:


maxnoe:
We decided that having separate offsets for drs4 and data is not needed, as this would change at the same time.

PR #119, opened 2021-11-12
Do not subtract an offset by default

maxnoe:
This changes how the global offset subtraction is done in
ctapipe_io_lst.
Instead of subtracting `offset` from both the waveform and the drs4
pedestal, we now rely on the offset being part of the drs4 pedestal.

This means, that when the drs4 pedestal calibration is used, no offset
needs to be given, making it possible to analyze runs with different
offsets without adapting the configuration.

### DIFF
--- a/ctapipe_io_lst/calibration.py
+++ b/ctapipe_io_lst/calibration.py
@@ -86,10 +86,12 @@ class LSTR0Corrections(TelescopeComponent):
     This calibrator exists in lstchain for testing and prototyping purposes.
     """
     offset = IntTelescopeParameter(
-        default_value=400,
+        default_value=0,
         help=(
-            'Define the offset of the baseline'
-            ', set to 0 to disable offset subtraction'
+            'Define the offset the offset of the baseline to be subtracted'
+            ' from the waveform *additionally* to the drs4 pedestal offset.'
+            ' Only set this to a value != 0 when not using the drs4 offset '
+            ' calibration.'
         )
     ).tag(config=True)
 
@@ -239,7 +241,9 @@ class LSTR0Corrections(TelescopeComponent):
             end = self.r1_sample_end.tel[tel_id]
             r1.waveform = r1.waveform[..., start:end]
 
-            r1.waveform -= self.offset.tel[tel_id]
+            if self.offset.tel[tel_id] != 0:
+                r1.waveform -= self.offset.tel[tel_id]
+
             mon = event.mon.tel[tel_id]
             if r1.selected_gain_channel is None:
                 r1.waveform[mon.pixel_status.hardware_failing_pixels] = 0.0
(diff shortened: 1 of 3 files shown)


ctapipe_io_lst/calibration.py
```diff
@@ -86,10 +86,12 @@ class LSTR0Corrections(TelescopeComponent):
     This calibrator exists in lstchain for testing and prototyping purposes.
     """
     offset = IntTelescopeParameter(
-        default_value=400,
+        default_value=0,
         help=(
-            'Define the offset of the baseline'
-            ', set to 0 to disable offset subtraction'
+            'Define the offset the offset of the baseline to be subtracted'
```
FrancaCassol, line 91: Typo in the first line.
Also the comment could be clearer, perhaps: "Define offset to be subtracted ADDITIONALLY to the drs4 baseline corrections (default)."

maxnoe: Thanks, will change

maxnoe: done

ctapipe_io_lst/calibration.py
```diff
@@ -88,8 +88,8 @@ class LSTR0Corrections(TelescopeComponent):
     offset = IntTelescopeParameter(
         default_value=0,
         help=(
-            'Define the offset the offset of the baseline to be subtracted'
-            ' from the waveform *additionally* to the drs4 pedestal offset.'
+            'Define offset to be subtracted from the waveform *additionally*'
+            ' to the drs4 pedestal offset.'
             ' Only set this to a value != 0 when not using the drs4 offset '
```
FrancaCassol, line 93: Is this true? We could want to subtract an additional offset also applying the drs4 correction offset. 
Otherwise, perhaps better to verify the apply_drs4_pedestal_correction flag before applying the offset and in the comment replace  *additionally* with *instead of*  (or similar)

maxnoe: I don't understand what you mean.

The default no is 0, so no offset is subtracted. We now also do not subtract an offset from the drs4 pedestal coefficents.

That means for the "normal" case of supplying a `drs4_pedestal_file`, everything should be correct, or not?

FrancaCassol: I refer to the phrase 'Only set this to a value != 0 when not using the drs4 offset'

maxnoe: Yes, I know, that's where you added the comment.

FrancaCassol: I mean, if we say that it is *additionally* to the drs4 pedestal offset, then it is not coherent  to ask to use it only if this is  not used

maxnoe: Yes it is. Because normally, the offset is included in the drs4 pedestal and you only need an `offset !=0` when not using the drs4_offset or when the offset of the drs4 run is different than the data run.

FrancaCassol: Hi @maxnoe, I continue to find the phrase "Only set this to a value != 0 when not using the drs4 offset " rather cryptical for a standard user and not necessary because the specification of *additionally* is enough to decide if to use or not this further offset

maxnoe: Is that cleared now?

FrancaCassol (approved):

PR #121, opened 2021-11-16
Implement and use spike subtraction by default, fixes #122

### LINKED ISSUE #122: spike correction should be ADC subtraction instead of the interpolation
Just let me report:
If predicted spikes are located at the pulse peak position, spike correction reduces the pulse height.
pulse_with_spike

As @tysaito2008 explained in the reco meeting, Tomohiko @toka2018 is studying the spike feature and implementing the new spike correction (ADC subtraction) into lstchain.

maxnoe:


maxnoe:
I tested this together with https://github.com/cta-observatory/cta-lstchain/pull/725 and it seems to work nicely.

Here are some example waveforms where the two methods (interpolation and subtraction) differ a lot (criterion for plotting was a difference of 500 in waveform.sum())


I think subtraction looks better in almost every case, here one prominent example:

spike_comparison




Especially pathological are interpolation cases where the interpolation actually makes the waveform higher, like this one:

spike_comparison


There are 941 waveforms (out of 53k * 1855 * 2) which differ by more than 500 counts in sub run `2965.0000`

moralejo:
Is this ready, @maxnoe ? I saw that the related https://github.com/cta-observatory/cta-lstchain/pull/725 has conflicts with the base branch. Would be nice to get this included in the next release.

maxnoe:
As written on slack, the first PR that needs to be merged is #108

rlopezcoto:
#108 was merged, can we move this to "Ready for review"?

maxnoe:
@rlopezcoto No, first we need the tool to create the needed new pedestal files in lstchain and add those to the test files.

ctapipe_io_lst/calibration.py
```diff
@@ -508,12 +532,9 @@ def time_lapse_corr(self, event, tel_id):
...
         """
-        Interpolates spike A & B.
-        Fill the R1 container.
-        Parameters
-        ----------
-        event : `ctapipe` event-container
-        tel_id : id of the telescope
+        Interpolate spikes at known positions from their neighboring values
+
+        Mutates the R1 waveform.
```
rlopezcoto, line 537: no `Parameters` or `Returns` given

maxnoe: It does not have a return value and the parameters are `event` and `tel_id`, do these really need explanations?

rlopezcoto: ok

rlopezcoto (approved):

FrancaCassol (approved):

PR #125, opened 2021-11-25
Set FF events not matching criteria to UNKNOWN

maxnoe:


moralejo (approved):
Looks fine to me. The test I included in the DL1 will now be redundant and never trigger, which means that the DL1 data check can no warn of this issue.
To spot the issue one should check the logs of r0_to_dl1, the same as for e.g. the UCTS jumps.

Do you think it would make sense to include these type of warnings among the other lst-specific informations filled by the event source, so that we can access them in an easier way than looking in the logs?

FrancaCassol:
Hi @maxnoe,
this solution is necessary for periods where the trigger flag does not work (like in the last months),
but in the case we know that the trigger flag works we would like to avoid to give priority  to the "by hand flag" which asks to give a correct FF intensity selection.  Could we find a way to give  preference to the trigger flag if wanted? (that hopefully will be the case soon)

moralejo:
Perhaps we should move the whole thing "up" and have the event source fill an "original" event_type flag and the heuristic one.

maxnoe:
> Do you think it would make sense to include these type of warnings among the other lst-specific informations filled by the event source, so that we can access them in an easier way than looking in the logs?

@moralejo the original one is still there in the `event.lst.tel[tel_id].evt` container. So you can always compare the trigger type that is filled into the `event.trigger` container with the information that decision was based on.

> Perhaps we should move the whole thing "up" and have the event source fill an "original" event_type flag and the heuristic one.
Then I can easily include the check in the "DL1" data checks.
For the same reason it would be nice also to have in the DL1 a record of other fixes, mainly the UCTS jumps.

Like I said, this information is still there. The only question is what is filled by default into the "native" ctapipe structure.

>  Could we find a way to give preference to the trigger flag if wanted? 

Yes, I will just add a boolean switch to not run the heuristic

maxnoe:
Going through the run @moralejo mentioned, the source warns about three events (using max events 30k, not all):

```
Found FF event that does not fulfill FF criteria: 15798
Found FF event that does not fulfill FF criteria: 16474
Found FF event that does not fulfill FF criteria: 26157
```

moralejo:
> > Perhaps we should move the whole thing "up" and have the event source fill an "original" event_type flag and the heuristic one.
> > Then I can easily include the check in the "DL1" data checks.
> > For the same reason it would be nice also to have in the DL1 a record of other fixes, mainly the UCTS jumps.
> 
> Like I said, this information is still there. The only question is what is filled by default into the "native" ctapipe structure.

Ok, thanks, then it is something for r0_to_dl1. But in order to include it in dl1 checks I'll have to pass it to dl1 files because they start from there.
 
And is there a similar way to notice the UCTS jumps corrected by the event source?

maxnoe:
> And is there a similar way to notice the UCTS jumps corrected by the event source?

trigger type UNKNOWN and event time unknown I think should be the event where a jump happened.

moralejo:
So could the event source pass the info on the jumps  somehow?

maxnoe:
Sure, we can add a bool if a jump was detected in that event to some of the custom containers

maxnoe:
@moralejo see https://github.com/cta-observatory/ctapipe_io_lst/pull/126

PR #126, opened 2021-11-25
Store jump info

maxnoe:


maxnoe:
@moralejo That's the info you wanted to have, right?

moralejo (approved):
Fine with me - will use it in the DL1 check as soon as this is released and bumped in lstchain.

PR #127, opened 2021-12-01
Unify functions for spike positions and spike positions just before the ROI

maxnoe:
Removes duplicated code by refactoring into a common function

### DIFF
--- a/ctapipe_io_lst/calibration.py
+++ b/ctapipe_io_lst/calibration.py
@@ -594,12 +594,16 @@ def get_spike_A_positions_base(current_first_cap, previous_first_cap, shift):
     for k in range(4):
         for base_position in base_positions:
             abspos = base_position + k * N_CAPACITORS_CHANNEL
+
             spike_A_position = (abspos - current_first_cap) % N_CAPACITORS_PIXEL
 
             # a spike affects the position itself and the two following slices
             # so we include also spikes two slices before the readout window
+            spike_A_position_shifted = spike_A_position - N_CAPACITORS_PIXEL
             if spike_A_position < N_SAMPLES:
                 positions.append(spike_A_position)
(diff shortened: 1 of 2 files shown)


ctapipe_io_lst/calibration.py
```diff
@@ -556,100 +556,104 @@ def interpolate_spike_A(waveform, position):
...
+    if last_capacitor % 2 != 0 or last_capacitor > last_in_first_half:
+        # bad trickery to get numba to compile an empty list with type int
+        # see https://numba.pydata.org/numba-doc/latest/user/troubleshoot.html
+        return [int(x) for x in range(0)]
+
+    # we have two cases for spikes that can occur in each of the 4 channels
+    base_positions = (
+        N_CAPACITORS_PIXEL - N_SAMPLES - previous_first_cap - shift,
+        previous_first_cap + N_SAMPLES - shift
+    )
```
SeiyaNozaki, line 591: I'd suggest using `last_capacitor` defined above to make it easier to understand.
```
best_positions = (
    N_CAPACITORS_PIXEL - last_capacitor - 2 - shift,
    last_capacitor - shift
)
```
and `shift` is 0, 1 for the current and old firmware respectively.
What do you think？

SeiyaNozaki (commented):
I think new lines to get spike positions with `shift` parameters are correct, and this PR seems fine.
I added a suggestion to make it easier to understand the spike conditions, but not a strong request.

ctapipe_io_lst/calibration.py
maxnoe: Perfect

maxnoe:
@jsitarek The code now produces the spike positions in the range [-2, 39]

SeiyaNozaki (approved):
It seems good, thanks!

PR #129, opened 2021-12-13
Add option to read interleaved pedestal event ids from file

maxnoe:


ctapipe_io_lst/__init__.py
```diff
@@ -762,3 +776,31 @@ def __len__(self):
...
         self.multi_file.close()
+
+    def read_pedestal_ids(self):
+        if self.pedestal_ids_path is not None:
+            t = read_table(self.pedestal_ids_path, '/interleaved_pedestal_ids')
+            Provenance().add_input_file(
+                self.pedestal_ids_path, role="InterleavedPedestalIDs"
+            )
+            self.pedestal_ids = set(t['event_id'])
+        else:
```
rlopezcoto, line 787: shouldn't we throw here a warning that no `pedestal_ids_path` was given?

maxnoe: No, I don't think so, since giving it is only required for a small number of old runs that don't properly tag the events, If I understood the problem correctly.

ctapipe_io_lst/tests/test_lsteventsource.py
```diff
@@ -243,3 +244,24 @@ def test_len():
...
     ) as source:
         assert len(source) == 10
+
+
+def test_pedestal_events(tmp_path):
+    from ctapipe_io_lst import LSTEventSource
+
+    path = tmp_path / 'pedestal_events.h5'
+    with tables.open_file(path, 'w') as f:
+        data = np.array([(2008, 5), (2008, 11)], dtype=[('obs_id', int), ('event_id', int)])
```
moralejo, line 254: Does not really matter for the test, but table created by pedestal finder does not contain obs_id. There is one pedestal_ids file per subrun, so it is not needed.

moralejo (approved):
Looks good to me, thanks!

ctapipe_io_lst/tests/test_lsteventsource.py
maxnoe: It does contain it since my latest commit....

PR #131, opened 2021-12-16
Fix biased r1 waveform after delta T correction

maxnoe:
An unintended side effect of relying on the drs4 offsets to also subtract the base offset of the waveform so that it does not have to be known is that now the waveform is already centered around 0 when applying the delta T correction.

The limit introduced here #95 thus now prevents any negative samples, biasing the pedestal estimation.

I see three options:

* Reintroduce the offset
* Remove the limit on the value that is subtracted from the waveform (done for now in this PR)
* Switch the order in which baseline and timelapse correction are applied, so that the waveform going into the timelapse correction is still around 400, not around 0.

Any opinions?

### DIFF
--- a/ctapipe_io_lst/calibration.py
+++ b/ctapipe_io_lst/calibration.py
@@ -968,8 +968,7 @@ def apply_timelapse_correction_pixel(
 
             # FIXME: Why only for values < 100 ms, negligible otherwise?
             if time_diff_ms < 100:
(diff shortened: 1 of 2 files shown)


FrancaCassol (approved):

FrancaCassol:
For now, this solution seems to me ok.
In future we will try to move to integer operations as in the EVB

PR #139, opened 2022-02-02
Fix for dying ucts

maxnoe:


ctapipe_io_lst/__init__.py
```diff
@@ -488,7 +488,10 @@ def fill_lst_event_container(self, array_event, zfits_event):
         """
         tel_id = self.tel_id
 
-        lst_evt = array_event.lst.tel[tel_id].evt
+        # create a fresh container so we are sure we have the invalid value
+        # markers in case on subsystem is going missing mid of run
+        lst_evt = LSTEventContainer()
+        array_event.lst.tel[1].evt = lst_evt
```
moralejo, line 494: [tel_id] ?

moralejo (changes requested):

ctapipe_io_lst/__init__.py
maxnoe: Yes, good catch

moralejo (approved):

PR #142, opened 2022-02-21
Read and add bending model corrections if next to drive report

maxnoe:


ctapipe_io_lst/pointing.py
```diff
@@ -55,10 +62,57 @@ def _read_drive_report(path):
                 'Az', 'azimuth_avg', 'azimuth_min', 'azimuth_max', 'azimuth_std',
                 'El', 'zenith_avg', 'zenith_min', 'zenith_max', 'zenith_std',
```
moralejo, line 63: If I get it right, this will work aso when the 'El' tag in the file is changed to the correct 'Zd'.  If so, we can just put here the new label, right? (to be confirmed with the Drive team)

ctapipe_io_lst/pointing.py
```diff
@@ -55,10 +62,57 @@ def _read_drive_report(path):
...
+            names=[
+                "weekday",
+                "month",
+                "day",
+                "time",
+                "year",
+                "unix_time",
+                "Az",
+                "azimuth_correction",
+                "El",
```
moralejo, line 110: Same as before - can we switch already to Zd?

ctapipe_io_lst/pointing.py
maxnoe: The name here does not play any role, I could also have added "_" or "foo" or someting else. It is ignored and not included in the output table (see `excluded_names`)

moralejo: Ok, so just give it the new name, or a dummy, to remove this wrong "elevation" reference which will soon disappear from the files as well.

moralejo (approved):

moralejo:
@maxnoe  can we merge this and make a new release so this becomes available?  We'd like to make a new release of lstchain as well.

maxnoe:
Ok with me

PR #149, opened 2022-06-02
Add focal length choice option for effective vs. nominal focal length

maxnoe:
uses effective by default


maxnoe:
New ctapipe will have both, we should directly update when released to that.

PR #153, opened 2022-07-28
ctapipe v0.15

vuillaut:


environment.yml
```diff
@@ -3,8 +3,8 @@ channels:
   - conda-forge
   - default
 dependencies:
-  - astropy>=4.2
-  - python=3.8 # nail the python version, so conda does not try upgrading / dowgrading
```
maxnoe, line 7: That = and comment is really important and must not be changed.

Isn't the comment clear enough?

vuillaut: It does not say why `python=3.8` is mandatory - and since the tests are passing with more recent versions, it was not clear to me why we should pin it.

maxnoe: It needs to be pinned to some version. Not to 3.8 specifically. But the sed command in the CI config uses the `python=` to insert the version used for the tests: 

https://github.com/cta-observatory/ctapipe_io_lst/blob/afac9ce3889d5e4c66042d08ba1b9135f3a4cdae/.github/workflows/ci.yml#L27

maxnoe: So if you want to change the default version, all good, but it should still be a fixed version.

maxnoe:
Closing in favor of #161

PR #156, opened 2022-09-28
Improve dtype handling in time_from_unix_tai_ns, improve test

maxnoe:
This fixes `time_from_unix_tai_ns` in case the input is not already an `np.uint64`, which was the case in `ctapipe_io_lst` but could easily not be the case if this function is imported somewhere else.

See this issue in `ctapipe_io_lst nectarcam` for more details:
https://github.com/cta-observatory/ctapipe_io_nectarcam/issues/24


maxnoe:
@FrancaCassol Could you review?

FrancaCassol (approved):

PR #157, opened 2022-10-26
Drop support for old dragon firmware

maxnoe:


rlopezcoto (commented):
Looks good, basically the code to read old firmware was removed, but shouldn't we leave the `LAST_RUN_WITH_OLD_FIRMWARE` defined and introduce an error whenever a file is open if you try to read a run previous to that one?

maxnoe:
> defined and introduce an error whenever a file is open if you try to read a run previous to that one?

No, the point to remove this code is actually, that the run id cannot be used since runs for LST2 data started over and always have the new firmware. So raising an error for old firmware runs would not solve the problem of analyzing LST2 camera data

rlopezcoto (approved):
then check run _and_ telescope number? 😄 
no, I think we are on the safe side, and anybody willing to access those runs (previous to the beginning of sky observations with LST-1), should be an expert that knows what is doing.

maxnoe:
> then check run and telescope number? smile

I'd like to, but:

```
In [2]: LSTEventSource("/pnfs/pic.es/data/cta/LST/LST2/Data/ZFITS/20220531/LST-2.1.Run00205.0000.fits.fz")
   ...: 
Out[2]: <ctapipe_io_lst.LSTEventSource at 0x7fe38b69cc40>

In [3]: s = _

In [4]: s.camera_config
Out[4]: 
CameraConfiguration(
    telescope_id=1
```

PR #166, opened 2023-01-06
Fix helpstrings of traitlets

maxnoe:
The commas are not wanted, we want the implicit string concatenation, not a tuple of two strings


maxnoe:
@FrancaCassol could you review?

FrancaCassol (approved):

PR #169, opened 2023-01-25
Use provision with micromamba in CI to fix conda installation

maxnoe:


maxnoe:
Same change as in lstchain in ctapipe

morcuended (approved):

PR #175, opened 2023-04-17
Handle missing TrackEnd by ending current tracking at time of TrackStart

maxnoe:


FrancaCassol (approved):

rlopezcoto:
merge @maxnoe ?

maxnoe:
@rlopezcoto I am not sure we want to.

The error this was supposed to address was a bug / crash of the program writing the drive reports, which has since been fixed and the files have been corrected.

I fear we might be missing new issues that just shouldn't happen if we introduce this behavior. I will close this for now.

PR #176, opened 2023-05-10
Allow and test with ctapipe 0.19

maxnoe:


rlopezcoto (approved):

rlopezcoto:
shall we go for v0.21 so it can also be updated on cta-lstchain and use ctapipe v0.19? I don't think we need to go for v0.19.1 for us...

maxnoe:
We should use it.

I would call this 0.20.1, since there are essentially no changes here.

rlopezcoto:
a bump in ctapipe version should imply another one in ctapipe_io_lst, not a bugfix one, right?

maxnoe:
Why? We did no changes to the code and we only *added* a supported version. 

We now support ctapipe 0.17 through 0.19

rlopezcoto:
and bumped the required version of ctapipe as well.
Anyways, I am also ok with v0.20.1

maxnoe:
> and bumped the required version of ctapipe as well.

No.

PR #188, opened 2023-06-09
Check # of gains in FF events. 

moralejo:
Check # of gains in FF events, and don't apply heuristic tag on 1-gain events.

### DIFF
--- a/src/ctapipe_io_lst/__init__.py
+++ b/src/ctapipe_io_lst/__init__.py
@@ -713,7 +713,7 @@ class LSTEventSource(EventSource):
         if waveform.ndim == 3:
             image = waveform[HIGH_GAIN].sum(axis=1)
         else:
-            return;
+            return
 
         in_range = (image >= self.min_flatfield_adc) & (image <= self.max_flatfield_adc)
         n_in_range = np.count_nonzero(in_range)
--- a/src/ctapipe_io_lst/__init__.py
+++ b/src/ctapipe_io_lst/__init__.py
@@ -708,27 +708,35 @@ class LSTEventSource(EventSource):
         '''
         tel_id = self.tel_id
         waveform = array_event.r1.tel[tel_id].waveform
-
-        # Tag only events with 2-gains waveforms: both gains are needed for calibration
+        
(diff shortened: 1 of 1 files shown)


src/ctapipe_io_lst/__init__.py
```diff
@@ -686,14 +686,12 @@ def fill_trigger_info(self, array_event):
...
         trigger.event_type = self._event_type_from_trigger_bits(trigger_bits)
-        
+
         if trigger.event_type == EventType.FLATFIELD:
-            waveform = array_event.r0.tel[tel_id].waveform
-            if waveform is None:
-                waveform = array_event.r1.tel[tel_id].waveform
-            if waveform.ndim == 2:
+            waveform = array_event.r1.tel[tel_id].waveform
+            if waveform is not None and waveform.ndim == 2:
```
moralejo, line 692: This is not identical to the previous code - just to make sure: we never need to check the dimensions of the R0 waveform? Is it always 3, in case R0 (and not R1) is filled?

moralejo: Anyway something is wrong with my earlier changes - test of lstchain_data_r0_to_dl1 on 12714.0100 does not work with --flatfield-heuristic (and it does work with the main branch)

moralejo: Ah, the problem was this:

https://github.com/cta-observatory/ctapipe_io_lst/blob/d0e5845f20fd32885ff54de333d806967f92e2dd/src/ctapipe_io_lst/__init__.py#L715-L716

Because now the event type of events trigger-tagged FF and not looking like FF is not set to UNKNOWN.

maxnoe: > This is not identical to the previous code - just to make sure: we never need to check the dimensions of the R0 waveform? Is it always 3, in case R0 (and not R1) is filled?

Yes, R0 is always two-gain (for LST)

maxnoe: Only the R1 container has the `selected_gain` field, this is why we put single-gain data into the R1 container even before the calibration...

PR #191, opened 2023-06-19
Add support for new CTA r1 data format (EVBv6)

maxnoe:
Rebased vs. main isntead of the 0.18.x branch.


moralejo (approved):

rlopezcoto:
@maxnoe is this ready to be merged to produce a new release?
I can take care of that of lstchain

maxnoe:
Not really tested on any data actually produced by the EVBv6, only on the artificial data I created in the unit tests, but I think we can go ahead and fix anything when we run into issues.

PR #198, opened 2023-09-22
Use importlib resource instead of pkg_resources

### LINKED ISSUE #197: pkg_resources is deprecated
Use importlib.resources instead.
although `importlib.resources.files` requires python >= 3.9 and currently we still support 3.8 as ctapipe v0.19.2 does

morcuended:
`importlib.resources.files` requires python >= 3.9 (required for next ctapipe v0.20.0)

closes #197

### DIFF
--- a/src/ctapipe_io_lst/__init__.py
+++ b/src/ctapipe_io_lst/__init__.py
@@ -2,54 +2,76 @@
 """
 EventSource for LSTCam protobuf-fits.fz-files.
 """
-from ctapipe.instrument.subarray import EarthLocation
 import logging
+import sys
+
 import numpy as np
 from astropy import units as u
-from importlib import resources
-from ctapipe.core import Provenance
-from ctapipe.instrument import (
-    ReflectorShape,
-    TelescopeDescription,
-    SubarrayDescription,
-    CameraDescription,
-    CameraReadout,
-    CameraGeometry,
-    OpticsDescription,
-    SizeType,
-)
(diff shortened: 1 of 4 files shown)


maxnoe:
> Pending as it requires python >= 3.9

There is a back-fill pypi module for versions < 3.9, which you should use like this:


```
import sys

if sys.version < (3, 9):
    from importlib_resources import ...
else:
    from importlib.resources import ...
```

And put the following into the requirements:

```
importlib_resources ; python_version < 3.9
```

morcuended:
Done @maxnoe

src/ctapipe_io_lst/__init__.py
```diff
@@ -2,54 +2,76 @@
...
 """
-from ctapipe.instrument.subarray import EarthLocation
 import logging
+import sys
+
 import numpy as np
 from astropy import units as u
-from importlib import resources
+
+if sys.version_info[:2] < (3, 9):
```
maxnoe, line 11: not really important, but the `[:2]` shouldn't be needed.

maxnoe:
@morcuended Sorry for not bringing this up earlier and making you do maybe unnecessary work, but in principle it would also be fine to just drop 3.8.

Next ctapipe version is already >= 3.9

morcuended:
Then I agree with dropping 3.8. In this PR or a separate one?

morcuended:
This was done somewhere else. Closing

PR #199, opened 2023-09-28
Ctar1 fixes

maxnoe:
Changes to make it fully work with EVBv6 data, not yet ready, will update as more test data is taken

Missing:

- [X] Read info on event-preprocessing done by EVB
- [ ] Apply only needed steps
- [ ] Missing module handling in code branch for new data format data (do we need this?)

### DIFF
--- a/src/ctapipe_io_lst/__init__.py
+++ b/src/ctapipe_io_lst/__init__.py
@@ -477,6 +477,7 @@ class LSTEventSource(EventSource):
 
         reordered_waveform = np.full((n_channels, N_PIXELS, n_samples), 0.0, dtype=np.float32)
         reordered_waveform[:, pixel_id_map] = waveform
+        waveform = reordered_waveform
 
         # FIXME, check using evb_preprocessing and make ctapipe support 2 gains
         if zfits_event.num_channels == 2:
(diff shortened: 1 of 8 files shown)


maxnoe:
This PR is already pretty large and is enough to get analysis working of data taken with EVBv6 without corrections applied or with manually specifying which corrections have already been applied.

Could you please review @FrancaCassol @moralejo or @SeiyaNozaki ?

I'll open another PR for the missing module and the automatic detection of calibration steps already performed by EVB

src/ctapipe_io_lst/__init__.py
```diff
@@ -437,40 +467,76 @@ def create_subarray(tel_id=1, reference_location=None):
 
     def fill_from_cta_r1(self, array_event, zfits_event):
         tel_id = self.tel_id
+        scale = self.data_stream.waveform_scale
+        offset = self.data_stream.waveform_offset
+        pixel_id_map = self.camera_config.pixel_id_map
+
+        # FIXME: missing modules / pixels
```
moralejo, line 474: So this will not work yet with  data with missing modules?

moralejo (approved):
Had a look at seems ok to me, but better if sb else checks it too...

src/ctapipe_io_lst/__init__.py
maxnoe: No, not yet, see my comment above. 

I'd like to separate that part into a second PR, with proper dummy test data and so on. This one will get pretty large if I want to include everything.

PR #201, opened 2023-11-13
Only create EventTimeCalculator in case of EVBv5 data

maxnoe:


moralejo:
@maxnoe  - question: does this mean we assume no issues with the timestamps ('UCTS jumps') in EvB6 data?

PR #208, opened 2024-01-15
Add option to correct event times by a fixed shift

maxnoe:


maxnoe:
@SeiyaNozaki what is the typical case for the WRS? I.e. would it be simpler to subtract or add the shift?

morcuended:
image
It is fine now

morcuended (approved):

PR #209, opened 2024-01-26
Fix definition of evb preprocessing flags, apply calibration only when needed

maxnoe:


src/ctapipe_io_lst/__init__.py
```diff
@@ -652,26 +651,36 @@ def _generator(self):
...
             # apply low level corrections
-            if self.apply_drs4_corrections:
-                self.r0_r1_calibrator.apply_drs4_corrections(array_event)
+            tdp_action = array_event.lst.tel[self.tel_id].evt.tdp_action
+            is_calibrated = False
+            if tdp_action is not None:
+                tdp_action = EVBPreprocessingFlag(int(tdp_action))
+                is_calibrated = EVBPreprocessingFlag.PE_CALIBRATION in tdp_action
 
+            if self.apply_drs4_corrections and not is_calibrated:
```
moralejo, line 660: is_calibrated refers to the conversion to p.e., right?
If apply_drs4_correction is True, shouldn't it always be the case that is_calibrated = False?

src/ctapipe_io_lst/evb_preprocessing.py
```diff
@@ -1,25 +1,45 @@
...
-    See EVB ICD section in
+    Notes
+    -----
+    This was supposed to be documented in the EVB ICD:
     https://edms.cern.ch/ui/file/2411710/2.6/LSTMST-ICD-20191206.pdf
+    But that document doest match the current EVB code.
+
+    As of 2024-01-26, there is a bug in EVB that the ttype_pattern and
+    tdp_action arrays are actually mixed up in the camera_configuration
+    object.
```
moralejo, line 19: Can this be explained a bit more? The meaning of both items, and the nature (and consequences) of the problem.

moralejo (approved):
A couple of minor comments

src/ctapipe_io_lst/__init__.py
maxnoe: `apply_drs4_corrections` is a user given option, `is_calibrated` is determined from the actual data.

So the logic here is mainly to simplify life for analyzers / osa that you can keep request drs4 calibration and give the drs4 calibration file, but the source will figure out if it is needed.

src/ctapipe_io_lst/evb_preprocessing.py
maxnoe: EVB just filled the real ttype_pattern into tdp_action and vice-versa. So the consequence is that if you want to know one, you actually have to look into the other field.

moralejo: Is that actually fixed now in the code anywhere? Or simply not yet used?

maxnoe: I checked again and realized I had not updated the `get_evb_preprocessings` function below. This is now done and I added a test.

As far as I know, this bug is not yet fixed in EVB.

moralejo (approved):

PR #212, opened 2024-03-05
Ignore missing TrackEnd when parsin TargetLog file by default

maxnoe:


morcuended (approved):
thanks Max! looks goods

PR #220, opened 2024-05-07
Fix test assert

maxnoe:
The assert was tautological...

### DIFF
--- a/src/ctapipe_io_lst/tests/test_lsteventsource.py
+++ b/src/ctapipe_io_lst/tests/test_lsteventsource.py
@@ -450,4 +450,4 @@ def test_arbitrary_filename(tmp_path):
         n_read = 0
         for _ in source:
             n_read += 1
-        assert n_read == n_read
+        assert n_read == 50


src/ctapipe_io_lst/tests/test_lsteventsource.py
```diff
@@ -450,4 +450,4 @@ def test_arbitrary_filename(tmp_path):
         n_read = 0
         for _ in source:
             n_read += 1
-        assert n_read == n_read
+        assert n_read == 50
```
morcuended, line 453: Is this the number of events in the file above (I guess newly added to the test dataset)?

morcuended (approved):
BTW, in the CI run before codecov failed (complaining something about tokens, something similar happened in lstmcpipe)

src/ctapipe_io_lst/tests/test_lsteventsource.py
maxnoe: It's not a new file, I just copied the existing test file into a file with a different name.

But yes, we know that file has 50 events

PR #230, opened 2024-11-08
support ctapipe 0.22

maxnoe:
We fixed scipy here to an issue in ctapipe, but this is fixed in the latest version of ctapipe and it doesn't make sense to limit it here in the package metadata.


morcuended (approved):
looks good

PR #235, opened 2024-11-25
Add new swat dtype definition

maxnoe:


morcuended:
I tried to produce the DRS4 baseline file with this branch version, but it failed, although the problem is unrelated.

```python
2024-11-26 09:28:59,980 ERROR [lstcam_calib.create_drs4_pedestal_file] (tool.run): Caught unexpected exception: Data has n_samples=40, expected 36. Applying offline drs4 corrections to data with border samples already removed by EVB is not supported.
```

PR #237, opened 2025-02-05
Test with ctapipe 0.23

maxnoe:


BastienLacave:
Hi,
I tested this branch with the lst_dl0_converter ans everything works on R0V data at least.
@moralejo It would be useful for us that this branch is merged and released so we solve the version incompatibility within the DL0 converter.
Thank you very much,
Bastien

maxnoe:
@moralejo this would be ready now

FrancaCassol (approved):

PR #238, opened 2025-03-07
Add reading of  calibration files in fits format

FrancaCassol:
This is necessary in order to move to fits output as requested by EvB

### DIFF
--- a/src/ctapipe_io_lst/calibration.py
+++ b/src/ctapipe_io_lst/calibration.py
@@ -406,9 +406,6 @@ class LSTR0Corrections(TelescopeComponent):
                     for col in row.keys():  
                         mon_data[key][col] = row[col]
             
-            # fits forget units if not defined in the container (to be corrected in ctapipe)
-            mon_data.calibration.time_correction = mon_data.calibration.time_correction*u.ns
-
             mon.tel[tel_id] = mon_data
 
         else:
--- a/src/ctapipe_io_lst/calibration.py
+++ b/src/ctapipe_io_lst/calibration.py
@@ -40,7 +40,8 @@ def to_native(data):
     if not data.dtype.isnative:
         data = data.byteswap()
         data = data.view(data.dtype.newbyteorder("="))
-        return data
+        
+    return data
 
 def get_first_capacitors_for_pixels(first_capacitor_id, expected_pixel_id=None):
     '''
--- a/src/ctapipe_io_lst/calibration.py
+++ b/src/ctapipe_io_lst/calibration.py
@@ -4,6 +4,7 @@ import numpy as np
 import astropy.units as u
 from numba import njit
 import tables
+from astropy.io import fits
 
 from ctapipe.core import TelescopeComponent
 from ctapipe.core.traits import (
@@ -14,6 +15,7 @@ from ctapipe.core.traits import (
 from ctapipe.calib.camera.gainselection import ThresholdGainSelector
 from ctapipe.containers import FlatFieldContainer, MonitoringCameraContainer, MonitoringContainer, PedestalContainer, PixelStatusContainer, WaveformCalibrationContainer
 from ctapipe.io import HDF5TableReader, read_table
+from astropy.table import QTable
 from traitlets import Enum
 
 from .compat import CTAPIPE_GE_0_21
@@ -34,6 +36,11 @@ __all__ = [
     'LSTR0Corrections',
 ]
 
+def to_native(data):
+    if not data.dtype.isnative:
+        data = data.byteswap()
+        data = data.view(data.dtype.newbyteorder("="))
+        return data
 
 def get_first_capacitors_for_pixels(first_capacitor_id, expected_pixel_id=None):
     '''
@@ -340,6 +347,7 @@ class LSTR0Corrections(TelescopeComponent):
             # time shift from flat fielding
             if self.mon_data is not None and self.add_calibration_timeshift:
                 time_corr = self.mon_data.tel[tel_id].calibration.time_correction
+                
                 # time_shift is subtracted in ctapipe,
                 # but time_correction should be added
                 if CTAPIPE_GE_0_21 or r1.selected_gain_channel is None:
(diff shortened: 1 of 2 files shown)


src/ctapipe_io_lst/calibration.py
```diff
@@ -34,6 +36,11 @@
     'LSTR0Corrections',
 ]
 
+def to_native(data):
+    if not data.dtype.isnative:
+        data = data.byteswap()
+        data = data.view(data.dtype.newbyteorder("="))
+        return data
```
maxnoe, line 43:
```suggestion
    return data
```
Otherwise, the function will return `None` in the case of the data already being native. 
You should add a test for this function.

src/ctapipe_io_lst/calibration.py
```diff
@@ -356,35 +364,77 @@ def _read_calibration_file(path):
...
-            with HDF5TableReader(path) as h5_table:
-                # read the calibration data
-                tel_id = int(base[4:])
-                mon.tel[tel_id] = MonitoringCameraContainer(
-                    calibration=next(h5_table.read(f'/{base}/calibration', WaveformCalibrationContainer)),
-                    pedestal=next(h5_table.read(f'/{base}/pedestal', PedestalContainer)),
-                    flatfield=next(h5_table.read(f'/{base}/flatfield', FlatFieldContainer)),
-                    pixel_status=next(h5_table.read(f"/{base}/pixel_status", PixelStatusContainer)),
-                )
+        if path.name.endswith(".h5"):
```
maxnoe, line 369: This is getting unwieldy. Please factor out the reading into functions, e.g.:

```
mon = read_calibration_h5(path)
```

src/ctapipe_io_lst/calibration.py
```diff
@@ -356,35 +364,77 @@ def _read_calibration_file(path):
...
+                "pedestal": PedestalContainer,
+                "pixel_status": PixelStatusContainer,
+            } 
+
+            mon_data = MonitoringCameraContainer()
+         
+            with fits.open(path) as f:
+                tel_id = f['PRIMARY'].header['TEL_ID']
+                for key in CALIB_CONTAINERS.keys():
+                    table = QTable.read(f, hdu=key)
```
maxnoe, line 402: I thought you store image hdus now?

src/ctapipe_io_lst/calibration.py
```diff
@@ -356,35 +364,77 @@ def _read_calibration_file(path):
...
+                    row = table[0]
+                    
+                    for col in row.keys():  
+                        mon_data[key][col] = row[col]
+            
+            # fits forget units if not defined in the container (to be corrected in ctapipe)
+            mon_data.calibration.time_correction = mon_data.calibration.time_correction*u.ns
```
maxnoe, line 409: What has to be corrected in ctapipe? You are not using any feature of ctapipe here

maxnoe:
You need to rebase or merge with main.

This also needs tests with the fits files produced by lstcam-calib.

src/ctapipe_io_lst/calibration.py
FrancaCassol: Non, these are BinTableHDU

src/ctapipe_io_lst/calibration.py
FrancaCassol: yes, the ctapipe containe when writing, it has not units defined

maxnoe: That doesn't matter here. You are reading from FITS.

src/ctapipe_io_lst/calibration.py
maxnoe: Why are we storing calibration in bintablehdu and everything else in image hdus?

maxnoe: I think this should be consistent, so that EVB and we can implement a single interface just using different extension and data names.

src/ctapipe_io_lst/calibration.py
FrancaCassol: no, as far as I understand the problem is  [here](https://github.com/cta-observatory/ctapipe/blob/7ef60cc5331f1880adccca93af26fbfd6b6c4ce8/src/ctapipe/containers.py#L1348 ), for fields with the units in the container (as [here](https://github.com/cta-observatory/ctapipe/blob/7ef60cc5331f1880adccca93af26fbfd6b6c4ce8/src/ctapipe/containers.py#L1330)) the unit is correctly propagated

src/ctapipe_io_lst/calibration.py
FrancaCassol: Now it is like that and Dirk did use dit already like that

FrancaCassol:
> You need to rebase or merge with main.

it tells it is already up to date

src/ctapipe_io_lst/tests/test_calib.py
```diff
@@ -25,6 +25,10 @@
 test_drs4_pedestal_path = calib_path / f'drs4_baseline/20200218/{calib_version}/drs4_pedestal.Run02005.0000.h5'
 test_time_calib_path = calib_path / f'drs4_time_sampling_from_FF/20191124/{calib_version}/time_calibration.Run01625.0000.h5'
 
+def test_to_native():
+    from ctapipe_io_lst.calibration import to_native
+
+    assert to_native(np.array([1, 2], dtype='>i4')).dtype=='int32'
```
maxnoe, line 31: This only tests the case where the array is not yet native and it also doens't verify that the values are actually correct.

maxnoe:
> > You need to rebase or merge with main.
> 
> it tells it is already up to date

Ah, sorry, right. The fix for the CI is not yet merged, this PR is needed, please review:
https://github.com/cta-observatory/ctapipe_io_lst/pull/237

src/ctapipe_io_lst/calibration.py
maxnoe: I will repeat myself, but: This branch is where we read a fits file written by lstcam_calib. If the unit is not there, it's our fault. Not ctapipe.

FrancaCassol:
> This also needs tests with the fits files produced by lstcam-calib.

@maxnoe, I need to add fits calibration and drs4 baseline files, but it seems that that the other test data are in lstchain, what do you suggest to do?

src/ctapipe_io_lst/tests/test_calib.py
```diff
@@ -172,18 +180,70 @@ def test_source_with_calibration():
...
+    with source:
+        for event in source:
+            assert event.r1.tel[1].waveform is not None
+            assert np.any(event.calibration.tel[1].dl1.time_shift != 0)
+
+@pytest.mark.parametrize("trigger_information", [False])
+@pytest.mark.parametrize(
+    "calib_config",
+    (
+        pytest.param(hdf5_calibration, id="fits"),
```
maxnoe, line 233: It doesn't make sense to parametrize a test if you only pass a single variant

src/ctapipe_io_lst/tests/test_calib.py
```diff
@@ -172,18 +180,70 @@ def test_source_with_calibration():
...
+    "calibration_path": test_calib_path_fits,
+}
+
+ 
 @pytest.mark.parametrize("trigger_information", [True, False])
-def test_source_with_all(trigger_information):
+@pytest.mark.parametrize(
+    "calib_config",
+    (
+        pytest.param(hdf5_calibration, id="hdf5"),
```
maxnoe, line 200: Why only hdf5 here? Why parametrize if you only pass one version?

src/ctapipe_io_lst/tests/test_calib.py
```diff
@@ -172,18 +180,70 @@ def test_source_with_calibration():
...
         'LSTEventSource': {
             'pointing_information': False,
-            'trigger_information': trigger_information,
+            'trigger_information': trigger_information, 
             'LSTR0Corrections': {
-                'drs4_pedestal_path': test_drs4_pedestal_path,
-                'drs4_time_calibration_path': test_time_calib_path,
-                'calibration_path': test_calib_path,
+                'drs4_pedestal_path': calib_config["drs4_baseline_path"],
+                'drs4_time_calibration_path': calib_config["drs4_time_path"],
```
maxnoe, line 212: If you use the same keys, you can just do `LSTR0Corrections: calib_config`

FrancaCassol:
I decided at the end to create only one test for all parameters

maxnoe (approved):

PR #249, opened 2026-03-06
Read in the event-and-pixel-wise pixel_time_shift information if avai…

moralejo:
Read in the event-and-pixel-wise pixel_time_shift information if available in the R1 stream

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,7 @@ dependencies:
   - astropy >=6.1,<8
   - python=3.11 # nail the python version, so conda does not try upgrading / dowgrading
   - ctapipe
-  - protozfits>=2.6
+  - protozfits>=2.11
   - eventio
   - corsikaio
   - zeromq
--- a/src/ctapipe_io_lst/__init__.py
+++ b/src/ctapipe_io_lst/__init__.py
@@ -567,14 +567,15 @@ class LSTEventSource(EventSource):
             stored_pixels = slice(None)  # all pixels stored
             n_pixels = zfits_event.num_pixels
         readout_shape = (n_channels, n_pixels)
-        reordered_pixel_time_shift = np.full((n_channels, N_PIXELS), 0.0, dtype=np.float32)
         if zfits_event.pixel_time_shift is not None:
+            reordered_pixel_time_shift = np.full((n_channels, N_PIXELS), 0.0, dtype=np.float32)
             raw_pixel_time_shift = zfits_event.pixel_time_shift.reshape(readout_shape)
             pixel_time_shift_ns = raw_pixel_time_shift.astype(np.float32) / np.float32(100.0)   # 10's of ps to ns
             pixel_id_map = self.camera_config.pixel_id_map
             reordered_pixel_time_shift[:, pixel_id_map[stored_pixels]] = pixel_time_shift_ns
-        pixel_time_shift = reordered_pixel_time_shift
-
+            pixel_time_shift = reordered_pixel_time_shift
+        else:
+            pixel_time_shift = None
         evt = LSTEventContainer(
             pixel_status=pixel_status,
             first_capacitor_id=zfits_event.first_cell_id,
@@ -685,7 +686,7 @@ class LSTEventSource(EventSource):
                 self.fill_r0r1_container(array_event, zfits_event)
                 self.fill_lst_event_container(array_event, zfits_event)
                 self.fill_trigger_info(array_event)
-
+                
             self.fill_mon_container(array_event, zfits_event)
 
             # apply correction before the rest, so corrected time is used e.g. for pointing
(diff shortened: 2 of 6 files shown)


src/ctapipe_io_lst/__init__.py
```diff
@@ -502,6 +502,13 @@ def fill_from_cta_r1(self, array_event, zfits_event):
         reordered_waveform[:, pixel_id_map[stored_pixels]] = waveform
         waveform = reordered_waveform
 
+        readout_shape = (n_channels, n_pixels)
+        raw_pixel_time_shift = zfits_event.pixel_time_shift.reshape(readout_shape)
+        pixel_time_shift_ns = raw_pixel_time_shift / 100   # 10's of ps to ns
```
maxnoe: I think this does integer computatation, at least in some versions of numpy, better be explicit:




```suggestion
        pixel_time_shift_ns = raw_pixel_time_shift.astype(np.float32) / np.float32(100.0)   # 10's of ps to ns
```

src/ctapipe_io_lst/__init__.py
```diff
@@ -560,12 +594,13 @@ def fill_lst_from_ctar1(self, zfits_event):
             evt.tdp_action = debug.tdp_action
 
             # unpack Dragon counters
-            counters = debug.counters.view(DRAGON_COUNTERS_DTYPE)
-            evt.pps_counter = counters['pps_counter']
-            evt.tenMHz_counter = counters['tenMHz_counter']
-            evt.event_counter = counters['event_counter']
-            evt.trigger_counter = counters['trigger_counter']
-            evt.local_clock_counter = counters['local_clock_counter']
+            if debug.counters is not None:
```
rlopezcoto: what is this debug doing exactly? aren't we in danger of not updating the counters if activated?

environment.yml
rlopezcoto, line 14: shouldn't we fix here the same numpy version as in lstchain to avoid problems with astropy incompatibilities?

src/ctapipe_io_lst/__init__.py
maxnoe: Debug is the debug data send by EVB. The only case where the counters are None when the debug block itself is present is if they are not present in the input data, which should never happen for real EVB data.

I will revert this here now, it was triggered by dummy data in the tests.

environment.yml
maxnoe: no. ctapipe_io_lst is compatible with newer versions of ctapipe than lstchain that are compatible with the recent astropy and numpy releases. This is needed for using `ctapipe_io_lst` directly with current ctapipe releases.

src/ctapipe_io_lst/__init__.py
maxnoe: done

moralejo:
@maxnoe @rlopezcoto can we merge this?

rlopezcoto:
I am happy with @maxnoe answers and modifications, so green light from my side

rlopezcoto (approved):

maxnoe:
The coverage complains are basically on all new code.

Can we add a test for this? Either by adding a small test file with the EVB pre-calibrated data, or by using the `dummy r1` here: 
https://github.com/cta-observatory/ctapipe_io_lst/blob/7efc480d5e1d0212e3ccf81e85902aca427f547d/src/ctapipe_io_lst/tests/test_cta_r1.py#L334-L355

maxnoe:
You can use:

```
❯ python -m protozfits.recompress_zfits --help
usage: recompress_zfits.py [-h] [-n N_EVENTS] [--rows-per-tile ROWS_PER_TILE]
                           [--block-size BLOCK_SIZE] [--overwrite] [--config CONFIG]
                           [--default-compression DEFAULT_COMPRESSION]
                           [-c COLUMN_COMPRESSION COLUMN_COMPRESSION] [--version]
                           inputfile outputfile

positional arguments:
  inputfile
  outputfile

options:
  -h, --help            show this help message and exit
  -n, --n-events N_EVENTS
                        If given, stop after n events
  --rows-per-tile ROWS_PER_TILE
                        Number of events per tile, if not given, same as inputfile
                        will be used
  --block-size BLOCK_SIZE
                        Compression block size in MB
  --overwrite           Overwrite output file without asking
  --config CONFIG       Path to yaml config file setting up compression options
  --default-compression DEFAULT_COMPRESSION
                        Default compression to be used, see adh-apis for supported
                        compressions
  -c, --column-compression COLUMN_COMPRESSION COLUMN_COMPRESSION
                        Compression to be used for a specific column
  --version             show program's version number and exit
```

to shrink the files by selecting the first 50 events.

moralejo:
> The coverage complains are basically on all new code.
> 
> Can we add a test for this? Either by adding a small test file with the EVB pre-calibrated data, or by using the `dummy r1` here:
> 
> https://github.com/cta-observatory/ctapipe_io_lst/blob/7efc480d5e1d0212e3ccf81e85902aca427f547d/src/ctapipe_io_lst/tests/test_cta_r1.py#L334-L355

Done, thanks! I'll merge.

PR #255, opened 2026-05-26
Pixel time correction set to have median=0 for each event

moralejo:
The correction applied to pixel times (for fully EvB_calibrated data) is now such that the median correction per event (for the full camera) is zero. In this way we modify as little as possible the times relative to the raw values, and they will still indicate (roughly) where the pulse is within the readout window. Since the corrections are correlated for many pixels, this means that the corrected times have now slightly larger fluctuations as "absolute times" w.r.t. the trigger moment. The std dev of the FF pulse times of any given pixel is a bit larger. The relative times among pixels (=what is relevant for the analysis) is of course unaffected. This can be modified later in Cat-B if desired (to recover the previous behaviour), but there is no actual advantage of doing so performance-wise.

### DIFF
--- a/src/ctapipe_io_lst/calibration.py
+++ b/src/ctapipe_io_lst/calibration.py
@@ -378,7 +378,7 @@ class LSTR0Corrections(TelescopeComponent):
                     # pixel_time_shift from lst.evt (or from r1 in the future)
                     # is (1, N_PIXELS), and contains the correction for the
                     # selected gain. On the other hand, time_shift in
-                    # event.calibration.tel[tel_id].dl1 is (2, N_PIXELS) no 
+                    # event.calibration.tel[tel_id].dl1 is (2, N_PIXELS) no
                     # matter whether the data is gain selected or not
                     time_shift = np.zeros((N_GAINS, N_PIXELS))
                     time_shift[r1.selected_gain_channel,
--- a/src/ctapipe_io_lst/calibration.py
+++ b/src/ctapipe_io_lst/calibration.py
@@ -4,6 +4,7 @@ import numpy as np
 import astropy.units as u
 from numba import njit
 import tables
+import logging
 from astropy.io import fits
 
 from ctapipe.core import TelescopeComponent
@@ -199,6 +200,9 @@ class LSTR0Corrections(TelescopeComponent):
             subarray=subarray, config=config, parent=parent, **kwargs
         )
 
+
+        self.log.setLevel(logging.INFO)
+
         self.mon_data = None
         self.last_readout_time = {}
         self.first_cap = {}
@@ -378,7 +382,11 @@ class LSTR0Corrections(TelescopeComponent):
                 # it later (in case e.g. a cat-A calibration file has
                 # been read in, and self.mon_data.tel[tel_id].calibration
                 # contains a non-zero time_correction value):
-                self.add_calibration_timeshift = False
+                if self.add_calibration_timeshift is True:
+                    self.log.info("NOTE: Pixel_time_shift provided by EVB in R1,"
+                                  " not adding time shift from calibration file"
+                    )
+                    self.add_calibration_timeshift = False
 
                 if r1.selected_gain_channel is None:
                     time_shift = lst.evt.pixel_time_shift # r1.pixel_time_shift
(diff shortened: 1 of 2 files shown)


src/ctapipe_io_lst/calibration.py
```diff
@@ -372,13 +372,28 @@ def fill_time_correction(self, event):
             # if r1.pixel_time_shift is not None:
             # Until pixel_time_shift is in R1CameraContainer... :
             if lst.evt.pixel_time_shift is not None:
+                # We assume that when pixel_time_shift is present
+                # in the input data (written out by EvB) it contains also
+                # inter-pixel time flatfielding. So we should not add
+                # it later (in case e.g. a cat-A calibration file has
+                # been read in, and self.mon_data.tel[tel_id].calibration
+                # contains a non-zero time_correction value):
+                self.add_calibration_timeshift = False
```
maxnoe: I'd add a `self.log.info("Setting ... to False")` here.

moralejo: Ah but wait this is event-wise. Can't do it like this.

maxnoe: Event-wise? Do you mean we need to support event by event EVB giving us the time shift or not? I would say that should be an error.

maxnoe:
```suggestion
            if self.add_calibration_timeshift is True:
                self.log.info("pixel_time_shift provided by EVB in R1, not adding time shift from calibration file")
                self.add_calibration_timeshift = False
```

moralejo: That should never be the case, indeed. I was just thinking on displaying the info upon initializing the event source, but  your solution is better, thanks!

src/ctapipe_io_lst/tests/test_lsteventsource.py
```diff
@@ -444,7 +444,14 @@ def test_evb_calibrated_data():
         read_events = 0
         for e in source:
             read_events += 1
-            assert np.all(e.calibration.tel[1].dl1.time_shift != 0)
+            time_shift = e.calibration.tel[1].dl1.time_shift
+            sg = e.r1.tel[1].selected_gain_channel
+            # Check that non-zero and different values are present for
+            # the selected channel(s):
+            if sg is None:
```
maxnoe: I find it a bit confusing that this is an unparametrized test with a single input file but there is an if/else condition here for something that should be fixed. What is the expected case here and what is dead code?

src/ctapipe_io_lst/calibration.py
```diff
@@ -398,6 +421,37 @@ def fill_time_correction(self, event):
...
+            # dependent (some pixels have a DRS4 time correction more
+            # correlated to the rest than others), so with this approach one
+            # can also distinguish 3 populations of pixels with std dev of
+            # FF times between 1.02 ans 1.10 ns. What really matters, which
+            # is the relative precision of the timing among pixels, is of
+            # course not affected by this correction, since it is a global
+            # shift per event.
+            #
+            if r1.selected_gain_channel is None:
+                time_shift = time_shift - np.nanmedian(time_shift)
```
maxnoe, line 447: Shouldn't the median here be per gain channel?

src/ctapipe_io_lst/tests/test_lsteventsource.py
moralejo: Right, I had in mind the tests we did with 2 gains and full calibration, which are part of the test data, but not used in this test.

src/ctapipe_io_lst/calibration.py
moralejo: I don't think so, we would break the relative calibration between HG and LG (which I hope is correct...). We should shift all  pixels by a fixed amount, no matter what gain was chosen.

src/ctapipe_io_lst/tests/test_lsteventsource.py
maxnoe: I thin that would actually be good. Parametrize the test with the different EVB pre-calibrated test data we have?

moralejo: Ok, I should upload the more recent test

maxnoe (approved):

Issue #245, opened 2025-06-26
Remove compat code for old ctapipe versions


maxnoe:
In #244, support for older versions of ctapipe was removed, but we still have some places in the code making checks on these ctapipe versions.

Can be removed now

PR #256, opened 2026-05-27
support latest ctapipe

tjarkmiener:
run latest and nightly ctapipe version in CI and allow failure for now; schedule trigger on 2:OO UTC

### DIFF
--- a/src/ctapipe_io_lst/__init__.py
+++ b/src/ctapipe_io_lst/__init__.py
@@ -25,8 +25,7 @@ from ctapipe.io.datalevels import DataLevel
 from ctapipe.core.traits import Bool, Float, Enum, Path
 from ctapipe.containers import (
     CoordinateFrameType, EventType, PointingMode, R0CameraContainer, R1CameraContainer,
-    SchedulingBlockContainer, ObservationBlockContainer, MonitoringContainer,
-    EventIndexContainer,
+    SchedulingBlockContainer, ObservationBlockContainer, EventIndexContainer,
 )
 from ctapipe.coordinates import CameraFrame
 
@@ -57,9 +56,9 @@ from .constants import (
 from .evb_preprocessing import get_processings_for_trigger_bits, EVBPreprocessingFlag
 from .compat import CTAPIPE_GE_0_20, CTAPIPE_GE_0_21, CTAPIPE_GE_0_27
 if CTAPIPE_GE_0_27:
-    from .containers import PixelStatusContainer, MonitoringCameraContainer
+    from .containers import PixelStatusContainer, MonitoringContainer, MonitoringCameraContainer
 else:
-    from ctapipe.containers import PixelStatusContainer, MonitoringCameraContainer
+    from ctapipe.containers import PixelStatusContainer, MonitoringContainer, MonitoringCameraContainer
 
 __all__ = [
     'LSTEventSource',
--- a/src/ctapipe_io_lst/__init__.py
+++ b/src/ctapipe_io_lst/__init__.py
@@ -657,7 +657,8 @@ class LSTEventSource(EventSource):
         # also add service container to the event section
 
         # initialize general monitoring container
-        mon = self.initialize_mon_container()
+        if not CTAPIPE_GE_0_27:
+            mon = self.initialize_mon_container()
 
         # loop on events
         for count, (_, zfits_event) in enumerate(self.multi_file):
@@ -669,14 +670,23 @@ class LSTEventSource(EventSource):
 
 
             # container for LST data
-            array_event = LSTArrayEventContainer(
-                count=count,
-                index=EventIndexContainer(
-                    obs_id=self.local_run_id,
-                    event_id=zfits_event.event_id,
-                ),
-                mon=mon,
-            )
+            if CTAPIPE_GE_0_27:
+                array_event = LSTArrayEventContainer(
+                    count=count,
+                    index=EventIndexContainer(
+                        obs_id=self.local_run_id,
+                        event_id=zfits_event.event_id,
+                    ),
+                )
+            else:
+                array_event = LSTArrayEventContainer(
+                    count=count,
+                    index=EventIndexContainer(
+                        obs_id=self.local_run_id,
+                        event_id=zfits_event.event_id,
+                    ),
+                    mon=mon,
+                )
             array_event.meta['input_url'] = self.input_url
             array_event.meta['max_events'] = self.max_events
             array_event.meta['origin'] = 'LSTCAM'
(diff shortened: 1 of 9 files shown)


src/ctapipe_io_lst/containers.py
```diff
@@ -145,8 +149,187 @@ class LSTContainer(Container):
...
+    )
+
+
 class LSTArrayEventContainer(ArrayEventContainer):
     """
     Data container including LST and monitoring information
     """
     lst = Field(default_factory=LSTContainer, description="LST specific Information")
+    # Patch
+    mon = Field(
```
maxnoe: Why do we add this here?

tjarkmiener: It was a quick patch, now removed @maxnoe

src/ctapipe_io_lst/__init__.py
```diff
@@ -747,8 +761,10 @@ def _generator(self):
                     relative_factor = np.empty((N_GAINS, N_PIXELS))
                     relative_factor[HIGH_GAIN] = self.r0_r1_calibrator.calib_scale_high_gain.tel[tel_id]
                     relative_factor[LOW_GAIN] = self.r0_r1_calibrator.calib_scale_low_gain.tel[tel_id]
-
-                    array_event.calibration.tel[tel_id].dl1.relative_factor = relative_factor
+                    if CTAPIPE_GE_0_27:
+                        array_event.monitoring.tel[tel_id].camera.coefficients.factor = relative_factor
```
tjarkmiener, line 755: @maxnoe I just noticed if we fill here this container the process tool will apply the calibration a second time. So we should just skip it here?

maxnoe: Yes, this the dl1 (Cat-B) calibration, so the initial calibration from R0 to R1 shouldn't be filled.

maxnoe: Wait: this is not filling an actual calibration, it's filling the `calib_scale_{low,high}_gain` values. So this is needed.

maxnoe:
Since you are editing now these parts and updating the ctapipe compatibility, you could remove the older checks (see https://github.com/cta-observatory/ctapipe_io_lst/issues/245)

src/ctapipe_io_lst/__init__.py
tjarkmiener: I'm a bit lost here @maxnoe It is still applied twice when using this LSTEventSource in the process tool?! Once filled in `calib_scale` we should set it back to defaults to avoid application in the process tool?

maxnoe: This is one of the last things done in the source, so it is not applied in `ctapipe_io_lst`, only filled.

src/ctapipe_io_lst/__init__.py
```diff
@@ -666,14 +669,23 @@ def _generator(self):
...
+                )
+            else:
+                array_event = LSTArrayEventContainer(
+                    count=count,
+                    index=EventIndexContainer(
+                        obs_id=self.local_run_id,
+                        event_id=zfits_event.event_id,
+                    ),
+                    mon=mon,
+                )
```
maxnoe: I'd just set mon afterwards to make the difference cleaner:


```suggestion
            array_event = LSTArrayEventContainer(
                count=count,
                index=EventIndexContainer(
                    obs_id=self.local_run_id,
                    event_id=zfits_event.event_id,
                ),
            )
            if not CTAPIPE_GE_0_27:
                array_event.mon = mon
```

src/ctapipe_io_lst/__init__.py
```diff
@@ -687,7 +699,8 @@ def _generator(self):
                 self.fill_lst_event_container(array_event, zfits_event)
                 self.fill_trigger_info(array_event)
 
-            self.fill_mon_container(array_event, zfits_event)
+            if not CTAPIPE_GE_0_27:
+                self.fill_mon_container(array_event, zfits_event)
```
maxnoe, line 694: Actually, just put it here

src/ctapipe_io_lst/tests/test_cta_r1.py
```diff
@@ -38,8 +38,7 @@
 subarray = LSTEventSource.create_subarray(tel_id=1)
 GEOMETRY = subarray.tel[1].camera.geometry
 pulse_shape = subarray.tel[1].camera.readout.reference_pulse_shape[0]
-if CTAPIPE_GE_0_21:
-    pulse_shape = pulse_shape[np.newaxis, ...]
+pulse_shape = pulse_shape[np.newaxis, ...]
```
maxnoe: Can be merged with the line above now

src/ctapipe_io_lst/tests/test_lsteventsource.py
```diff
@@ -332,15 +339,21 @@ def test_pointing_info():
...
+            if CTAPIPE_GE_0_27:
+                assert u.isclose(e.monitoring.pointing.array_ra, expected_ra)
+                assert u.isclose(e.monitoring.pointing.array_dec, expected_dec)
+                assert u.isclose(e.monitoring.tel[1].pointing.altitude.to(u.deg), expected_alt, rtol=1e-2)
+                assert u.isclose(e.monitoring.tel[1].pointing.azimuth.to(u.deg), expected_az, rtol=1e-2)
+            else:
+                assert u.isclose(e.pointing.array_ra, expected_ra)
+                assert u.isclose(e.pointing.array_dec, expected_dec)
+                assert u.isclose(e.pointing.tel[1].altitude.to(u.deg), expected_alt, rtol=1e-2)
+                assert u.isclose(e.pointing.tel[1].azimuth.to(u.deg), expected_az, rtol=1e-2)
```
maxnoe: I'd try to minimize the difference between the versions to make it easier to change / see where differences are:


```suggestion
            if CTAPIPE_GE_0_27:
                array_pointing = e.monitoring.pointing
                tel_pointing = e.monitoring.tel[1].pointing
            else:
                array_pointing = e.pointing
                tel_pointing = e.tel[1].pointing                

            assert u.isclose(array_pointing.array_ra, expected_ra)
            assert u.isclose(array_pointing.array_dec, expected_dec)
            assert u.isclose(tel_pointing.altitude.to(u.deg), expected_alt, rtol=1e-2)
            assert u.isclose(tel_pointing.azimuth.to(u.deg), expected_az, rtol=1e-2)
```

maxnoe:
Given that we plan to pull the trigger on CEP-002 soon, I'd prefer to keep the upper version bound here and be explicit about the versions we test, e.g. replace `latest` with 0.30 and keep the upper version bound.

maxnoe (approved):

maxnoe:
Thanks a lot!

PR #257, opened 2026-05-28
Change TEST_DATA_URL to new endpoint

moralejo:
Update TEST_DATA_URL to the new base URL.

### DIFF
--- a/download_test_data.sh
+++ b/download_test_data.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-TEST_DATA_URL=${TEST_DATA_URL:-https://old.lst.iac.es/lstchain-testfiles/}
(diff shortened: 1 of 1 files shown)


maxnoe (approved):

tjarkmiener:
Can we use this test data also for the DL0Converter repo? @maxnoe

moralejo:
> Can we use this test data also for the DL0Converter repo? @maxnoe

I think so. It does not seem to have any  test data so far, right? @TjarkMiener

tjarkmiener:
> I think so. It does not seem to have any test data so far, right? @TjarkMiener

No, @BastienLacave was using private LSTData for this local unittests. Having something set up similar to here would be nice (I think we only need some calibration files and a corresponding data file)

PR #265, opened 2026-07-02
Read in and apply timelapse DRS4 baseline corrections from a fits file.

moralejo:
Allows different correction parameters for different batches of pixels.

### DIFF
--- a/src/ctapipe_io_lst/calibration.py
+++ b/src/ctapipe_io_lst/calibration.py
@@ -136,6 +136,17 @@ class LSTR0Corrections(TelescopeComponent):
         ),
     ).tag(config=True)
 
+    drs4_timelapse_path = TelescopeParameter(
+        trait=Path(exists=True, directory_ok=False, allow_none=True),
+        allow_none=True,
+        default_value=None,
+        help=(
+            'Path to the LST pedestal timelapse baseline correction file'
+            'If not provided, the default LST1 correction will be applied'
+            'if apply_timelapse_correction=True'
+        ),
+    ).tag(config=True)
+
     calibration_path = Path(
         None, exists=True, directory_ok=False, allow_none=True,
         help='Path to LST calibration file',
@@ -221,6 +232,7 @@ class LSTR0Corrections(TelescopeComponent):
         self.first_cap_old = {}
         self.fbn = {}
         self.fan = {}
+        self.timelapse_correction_params = {}
 
         for tel_id in self.subarray.tel:
             shape = (N_GAINS, N_PIXELS, N_CAPACITORS_PIXEL)
@@ -238,16 +250,23 @@ class LSTR0Corrections(TelescopeComponent):
         else:
             self.gain_selector = None
 
+        if self.apply_timelapse_correction:
+            for tel_id in self.subarray.tel:
+                tlapse_file = self.drs4_timelapse_path.tel[tel_id]
+                if tlapse_file is not None:
+                    self.timelapse_correction_params[tel_id] = self._read_timelapse_file(tlapse_file)
+
         if self.calibration_path is not None:
             self.mon_data = self._read_calibration_file(self.calibration_path)
         else:
             # In case no calibration file has been provided (like e.g. when processing calibrated R1)
             # we have to create the container here:
             mon = MonitoringContainer()
-            mon_camera = MonitoringCameraContainer()
-            mon_camera.calibration.time_correction = np.zeros((N_GAINS, N_PIXELS)) * u.ns
-            mon_camera.calibration.unusable_pixels = np.zeros((N_GAINS, N_PIXELS), dtype='bool')
-            mon.tel[tel_id] = mon_camera
+            for tel_id in self.subarray.tel:
+                mon_camera = MonitoringCameraContainer()
+                mon_camera.calibration.time_correction = np.zeros((N_GAINS, N_PIXELS)) * u.ns
+                mon_camera.calibration.unusable_pixels = np.zeros((N_GAINS, N_PIXELS), dtype='bool')
+                mon.tel[tel_id] = mon_camera
             self.mon_data = mon
             
     def apply_drs4_corrections(self, event: LSTArrayEventContainer):
@@ -472,6 +491,22 @@ class LSTR0Corrections(TelescopeComponent):
             else:
                 event.calibration.tel[tel_id].dl1.time_shift = time_shift
 
+    @staticmethod
+    def _read_timelapse_file(path):
+        """
+        Read the timelapse baseline correction parameters from fits timelapse file
+        Returns a (3, ngains, npixels) array, the first index stands for the 3
+        necessary parameters (scale, exponent and t0). The baseline correction
+        is scale * ((delta_t / t0)**-exponent - 1)  (to be subtracted from the
+        given DRS4 cell baseline)
+        """
+        with fits.open(path) as f:
+           pixel_batch = f['PIXEL_BATCH'].data # (ngains, npixels)
+           scale = f['SCALE'].data[pixel_batch]       # (ngains, npixels)
+           exponent = f['EXPONENT'].data[pixel_batch] # (ngains, npixels)
+           t0 = f['T0'].data[pixel_batch]             # (ngains, npixels)
+
+        return np.array([scale, exponent, t0])
 
     @staticmethod
     def _read_calibration_file(path):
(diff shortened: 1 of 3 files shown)


src/ctapipe_io_lst/calibration.py
```diff
@@ -238,16 +250,23 @@ def __init__(self, subarray, config=None, parent=None, **kwargs):
         else:
             self.gain_selector = None
 
+        if self.apply_timelapse_correction:
+            for tel_id in self.subarray.tel:
+                tlapse_file = self.drs4_timelapse_path.tel[tel_id]
```
FrancaCassol, line 255: I see that you consider the possibility of several telescopes, but each the time_lapse file concerns only one telescope. Also, I do not see where you associate the telescope subarray to the drs4_timelapse_path (tough perhaps I miss it)

moralejo: I think the assignment is done automatically if we ever put more than one telescope in the same config (each with its file). anyway, I wrote it like that for consistency with the rest, this part will always be executed telescope-wise.

FrancaCassol: I honestly   do not see the possibility to have more of one telescope in the present code version (see the [init function](https://github.com/cta-observatory/ctapipe_io_lst/blob/820947c7535528684d080a8bc6df7fe03d9cebb1/src/ctapipe_io_lst/__init__.py#L343) ) to which "rest" of the code do you refers?

moralejo: I referred to all members of class LSTR0Corrections which are *TelescopeParameter 
I agree this is going to be executed telescope-wise, but this is also possible with the current implementation.

maxnoe: This mirrors the current implementation of all other calibration files, so I would change it

src/ctapipe_io_lst/calibration.py
```diff
@@ -472,6 +491,22 @@ def fill_time_correction(self, event):
             else:
                 event.calibration.tel[tel_id].dl1.time_shift = time_shift
 
+    @staticmethod
+    def _read_timelapse_file(path):
```
maxnoe, line 500: Do we really want to blow up the array to a value per gain, pixel? I'm worried what this does to the CPU cache when applying the correction.

Why not keep the batch array and the values per batch?

src/ctapipe_io_lst/calibration.py
```diff
@@ -1109,9 +1151,9 @@ def apply_timelapse_correction_pixel(
             time_diff = time_now - last_readout_time_cap
             time_diff_ms = time_diff / CLOCK_FREQUENCY_KHZ
 
-            # FIXME: Why only for values < 100 ms, negligible otherwise?
+            # Only for values < 100 ms, negligible otherwise
             if time_diff_ms < 100:
```
maxnoe: This line needs to be adjusted: for the new function, the cutoff is not the fixed 100 but the scale parameter.

src/ctapipe_io_lst/calibration.py
```diff
@@ -1235,11 +1287,19 @@ def apply_timelapse_correction_gain_selected(
             pixel_id = expected_pixels_id[pixel_index]
             gain = selected_gain_channel[pixel_id]
 
+            if tlapse_params is None:
```
maxnoe: I'm a bit worried what this branching inside of the compiled function does to performance.

Is the new method much slower than before? Both when using the old approach or the new?

I think we best would be two separate functions for the two approaches, not branching inside.

src/ctapipe_io_lst/calibration.py
```diff
@@ -1254,19 +1314,28 @@ def apply_timelapse_correction_gain_selected(
 
 
 @njit(cache=True)
-def ped_time(timediff):
+def ped_time(timediff, params):
```
maxnoe, line 1326: Same here, I'd use two completely separate functions to avoid the branching.

src/ctapipe_io_lst/calibration.py
```diff
@@ -1254,19 +1314,28 @@ def apply_timelapse_correction_gain_selected(
...
+        # return 27.33 * np.power(timediff, -0.24) - 10.4
+        # new values at 20 degC, provided by Yokiho Kobayashi 2/3/2020
+        # see also Yokiho's talk in https://indico.cta-observatory.org/event/2664/
+        return 32.99 * timediff**(-0.22) - 11.9
+    else:
+        # Values from file (pixels with DRS4 from different batches)
+        correction = params[0] * ((timediff / params[2])**-params[1] - 1)
+        # The parametrization is valid until it becomes negative. Beyond that
+        # (long timediffs), correction is 0
+        correction = max(correction, 0)
```
maxnoe: This should not be needed, the function should only be reached if timediff <= t0

src/ctapipe_io_lst/calibration.py
moralejo: Do you mean the memory taken by the parameters? I assumed that in absolute terms 3 * 2 * 1855 floats is still negligible.

maxnoe: In memory yes, but maybe not in the CPU cache, so it might affect the performance of applying the corrections. I didn't check though.

src/ctapipe_io_lst/calibration.py
moralejo: The old function can be expressed using the new function's parameters, I will just pass the right parameters to the function and remove the branching inside it.

src/ctapipe_io_lst/calibration.py
```diff
@@ -1170,11 +1217,15 @@ def apply_timelapse_correction(
                 pixel_index = module * N_PIXELS_MODULE + pixel_in_module
                 pixel_id = expected_pixels_id[pixel_index]
 
+                # Corrections parameters for this pixel & gain:
+                tlp = tlapse_params[:, gain, pixel_id]
```
maxnoe: Inverting the hierarchy makes more sense then in this case, right? Because you access the parameters together: `tlapse_param[gain, pixel_id]`

maxnoe: this makes sure the values are close together in memory as they are used

src/ctapipe_io_lst/calibration.py
```diff
@@ -691,6 +726,16 @@ def time_lapse_corr(self, event, tel_id):
         # The old readout (before 2019/11/05) is shifted by 1 cell.
         run_id = event.lst.tel[tel_id].svc.configuration_id
 
+        if tel_id in self.timelapse_correction_params:
+            tlc = self.timelapse_correction_params[tel_id]
+        else:
+            # if not provided, use default values for LST1
+            # Values at 20 degC, provided by Yukiho Kobayashi 2/3/2020
+            # see also Yukiho's talk in https://indico.cta-observatory.org/event/2664/
+            tlc = np.array([11.9, 0.22, 103.012])
```
maxnoe: Then this would just be `np.full((N_GAINS, N_PIXELS, 3), tlc)`

src/ctapipe_io_lst/calibration.py
moralejo: I will anyway change it to do it by batch, as you suggested.

src/ctapipe_io_lst/calibration.py
```diff
@@ -238,16 +255,23 @@ def __init__(self, subarray, config=None, parent=None, **kwargs):
         else:
             self.gain_selector = None
 
+        if self.apply_timelapse_correction:
+            for tel_id in self.subarray.tel:
+                tlapse_file = self.drs4_timelapse_path.tel[tel_id]
+                if tlapse_file is not None:
+                    self.drs4_batch[tel_id], self.timelapse_correction_params[tel_id] = self._read_timelapse_file(tlapse_file)
```
maxnoe: I think we should init the old values here instead of each time it is called.

maxnoe: I.e. add an

```
else:
    self.drs4_batch[tel_id], self.timelapse_correction_params[tel_id] = self._get_default_time_params()
```

and maybe raise an error if it is not `tel_id == 1`.

maxnoe:
A general comment is that I think we should start thining about moving the DRS4 correction code to `lstcam_calib`, as in future,  I think we will read standard DL0 from calibration runs, using the standard `ctapipe_io_zfits`. But then the corrections still need to be performed on the fly to compute the calibration files.

PR #266, opened 2026-08-07
Avoid warning about missing reference metadata when loading pulse shape file

maxnoe:


moralejo (approved):
Thanks, those warnings were quite annoying

